PR #14, opened 2017-06-15
Allow doc comment to set help text

### LINKED ISSUE #13: Consider using field doc comment as help text by default
When no `help = "Foo"` argument is given in the `structopt` attribute on a field, you could use the field's doc comment if present. This might make the code a bit clearer while also reducing duplicating the text.

killercup:
This treats doc comments as `help`/`about` attributes, which may lead to concatenating the original help text with the doc comment.

I didn't add any tests except for changing the 'basic' example and checking the CLI output myself. If there is anything you want me to add in that regard, let me know :)

fix #13

### DIFF
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -172,7 +172,7 @@ fn extract_attrs<'a>(attrs: &'a [Attribute], attr_source: AttrSource) -> Box<Ite
     Box::new(doc_comments.chain(settings_attrs))
 }
 
-fn from_attr_or_env<'a>(attrs: &[(Ident, Lit)], key: &str, env: &str) -> Lit {
+fn from_attr_or_env(attrs: &[(Ident, Lit)], key: &str, env: &str) -> Lit {
     let default = std::env::var(env).unwrap_or("".into());
     attrs.iter()
         .filter(|&&(ref i, _)| i.as_ref() == key)
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -169,20 +169,22 @@ fn extract_attrs<'a>(attrs: &'a [Attribute], attr_source: AttrSource) -> Box<Ite
             }
         });
 
-    Box::new(settings_attrs.chain(doc_comments))
+    Box::new(doc_comments.chain(settings_attrs))
 }
 
 fn from_attr_or_env<'a>(attrs: &[(Ident, Lit)], key: &str, env: &str) -> Lit {
     let default = std::env::var(env).unwrap_or("".into());
     attrs.iter()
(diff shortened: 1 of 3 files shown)


structopt-derive/src/lib.rs
```diff
@@ -129,31 +129,62 @@ fn sub_type(t: &syn::Ty) -> Option<&syn::Ty> {
     }
 }
 
-fn extract_attrs<'a>(attrs: &'a [Attribute]) -> Box<Iterator<Item = (&'a Ident, &'a Lit)> + 'a> {
-    let iter = attrs.iter()
+#[derive(Debug, Clone, Copy)]
+enum AttrSource { Struct, Field, }
+
+fn extract_attrs<'a>(attrs: &'a [Attribute], attr_source: AttrSource) -> Box<Iterator<Item = (Ident, Lit)> + 'a> {
```
killercup: I was pretty close to making this a `Box<Iterator<Item = (Cow<'a, Ident>, Cow<'a, Lit>)>` but luckily I noticed early enough that this was derive macro whose performance will probably never matter 😅

TeXitoi:
Great work, thanks!

I'd prefer that if help is provided, help is used, else the docstring is used. But it looks like the last seen will be used, thus better to revert the chain. To be sure of the result, can you add an example in `tests/` with about and a docstring on the struct, and with help and a docstring on a field, and testing with `ArgsMatches::usage()` ?

Thanks.

killercup:
Alright, help/about is now preferred if the doc string is also set, and I added some tests comparing the help output that clap's `write_long_help` gives me.

structopt-derive/src/lib.rs
```diff
@@ -129,31 +129,64 @@ fn sub_type(t: &syn::Ty) -> Option<&syn::Ty> {
...
+            } else {
+                None
+            }
+        });
+
+    Box::new(doc_comments.chain(settings_attrs))
 }
 
-fn from_attr_or_env(attrs: &[(&Ident, &Lit)], key: &str, env: &str) -> Lit {
+fn from_attr_or_env<'a>(attrs: &[(Ident, Lit)], key: &str, env: &str) -> Lit {
```
TeXitoi: Useless lifetime parameter

killercup: Fixed

TeXitoi:
Thanks! I'll update crates.io after eating.

killercup:
Thanks! Bon appétit!

TeXitoi:
Published

killercup:
I think you just published a new version of structopt but not structopt-derive?

TeXitoi:
oups, should be fixed

Issue #1, opened 2017-02-07
Support subcommand


TeXitoi:
The idea would be to create StructOpt struct for each subcommand, and then dispatching on an enum on the principal StructOpt.

ErichDonGubler:
So, something like this?

```rust
#[derive(StructOpt, Debug)]
#[structopt(name = "git", about = "Because git is great!")]
enum GitCommand {
    #[structopt(name = "fetch")]
    Fetch {
        #[structopt(help = "Remote to fetch from", default_value = "None")]
        remote: Option<&str>, // Ignoring lifetime here
        #[structopt(help = "Branch to fetch from", default_value = "None")]
        branch: Option<&str> // Ignoring lifetime here
    },
    #[structopt(name = "commit")]
    Commit {
        #[structopt(short = "a", long = "all", help = "Adds all tracked files to the index before commit", default_value = "false")]
        all: bool,
        #[structopt(short = "p", long = "patch", help = "Use interactive patching interface for staging", default_value = "false")]
        patch: bool,
        #[structopt(short = "m", long = "message", help = "Add a commit message via CLI args instead of using $EDITOR", default_value = "None")]
        message: Option<&str> // Ignoring lifetime here
    }
}

fn main () {
    match Opt::from_args() {
        Fetch { remote, branch } => {
            //...
        },
        Commit { all, patch, message } => {
            //...
        }
    }
}
```

TeXitoi:
The idea was:
```rust
#[derive(StructOpt)]
#[structopt(name = "git", about = "Because git is great!")]
struct App {
    #[structopt(subcommand)]
    subcommand: SubCommand,
}
#[derive(StructOptSubCommand)]
enum SubCommand {
     #[structopt(name = "fetch")]
    Fetch(Fetch),
    #[structopt(name = "commit")]
    Commit(Commit),
}
#[derive(StructOpt)]
struct Fetch {
    #[structopt(help = "Remote to fetch from")]
    remote: Option<String>,
    #[structopt(help = "Branch to fetch from")]
    branch: Option<String>,
}
#[derive(StructOpt)]
struct Commit {
    #[structopt(short = "a", long = "all", help = "Adds all tracked files to the index before commit")]
    all: bool,
    #[structopt(short = "p", long = "patch", help = "Use interactive patching interface for staging")]
    patch: bool,
    #[structopt(short = "m", long = "message", help = "Add a commit message via CLI args instead of using $EDITOR")]
    message: Option<String>,
}
[...]

PR #17, opened 2017-07-03
Add support for subcommands

williamyaoh:
See #1 for detailed discussion of this feature.

Adds the ability to use `#[derive(StructOpt)]` on enums, thereby allowing the creation/easy parsing of `clap::SubCommand`s.

**This is not a breaking change.**

## Things that can still be improved:
+ [X] Fix subcommand names that are hyphenated
+ [ ] Automatically downcase the names of subcommands for help messages (this could also apply to `#[derive(StructOpt)]` on structs or to changing argument names from `_` to `-`)
+ [x] Add the ability to use `about` attribute/doc comments on subcommand variants for help messages
+ [x] Allow other enum variant types than curly-bracketed ones (e.g. "blank" subcommands)

## Other stuff

There are some commits that don't technically have anything to do with adding subcommands in:

+ Add documentation for the PR that added doc comments for help messages a few weeks ago
+ Add a link from the `structopt` documentation to the `structopt-derive` documentation
+ Bump versions of `structopt` and `structopt-derive` to 0.1.0
+ Edit the `structopt-derive` documentation for clarity

Detailed documentation has been added to the `structopt-derive` crate specifying how to use the new subcommand functionality. Reproduced here:

## Subcomamnds

Some applications, like `git`, support "subcommands;" an extra command that
is used to differentiate what the application should do. With `git`, these
would be `add`, `init`, `fetch`, `commit`, for a few examples.

`clap` has this functionality, so `structopt` supports this through enums:

```rust
#[derive(StructOpt)]
#[structopt(name = "git", about = "the stupid content tracker")]
enum Git {
    #[structopt(name = "add")]
    Add {
        #[structopt(short = "i")]
        interactive: bool,
        #[structopt(short = "p")]
        patch: bool,
        files: Vec<String>
    },
    #[structopt(name = "fetch")]
    Fetch {
        #[structopt(long = "dry-run")]
        dry_run: bool,
        #[structopt(long = "all")]
        all: bool,
        repository: Option<String>
    },
    #[structopt(name = "commit")]
    Commit {
        #[structopt(short = "m")]
        message: Option<String>,
        #[structopt(short = "a")]
        all: bool
    }
}
```

Using `derive(StructOpt)` on an enum instead of a struct will produce
a `clap::App` that only takes subcommands. So `git add`, `git fetch`,
and `git commit` would be commands allowed for the above example.

`structopt` also provides support for applications where certain flags
need to apply to all subcommands, as well as nested subcommands:

```rust
#[derive(StructOpt)]
#[structopt(name = "make-cookie")]
struct MakeCookie {
    #[structopt(name = "supervisor", default_value = "Puck", required = false, long = "supervisor")]
    supervising_faerie: String,
    #[structopt(name = "tree")]
    /// The faerie tree this cookie is being made in.
    tree: Option<String>,
    #[structopt(subcommand)]  // Note that we mark a field as a subcommand
    cmd: Command
}

#[derive(StructOpt)]
enum Command {
    #[structopt(name = "pound")]
    /// Pound acorns into flour for cookie dough.
    Pound {
        acorns: u32
    },
    #[structopt(name = "sparkle")]
    /// Add magical sparkles -- the secret ingredient!
    Sparkle {
        #[structopt(short = "m")]
        magicality: u64,
        #[structopt(short = "c")]
        color: String
    },
    #[structopt(name = "finish")]
    Finish {
        #[structopt(short = "t")]
        time: u32,
        #[structopt(subcommand)]  // Note that we mark a field as a subcommand
        type: FinishType
    }
}

#[derive(StructOpt)]
enum FinishType {
    #[structopt(name = "glaze")]
    Glaze {
        applications: u32
    },
    #[structopt(name = "powder")]
    Powder {
        flavor: String,
        dips: u32
    }
}
```

Marking a field with `structopt(subcommand)` will add the subcommands of the
designated enum to the current `clap::App`. The designated enum *must* also
be derived `StructOpt`. So the above example would take the following
commands:

+ `make-cookie pound 50`
+ `make-cookie sparkle -mmm --color "green"`
+ `make-cookie finish 130 glaze 3`

### Optional subcommands

A nested subcommand can be marked optional:

```rust
#[derive(StructOpt)]
#[structopt(name = "foo")]
struct Foo {
    file: String,
    #[structopt(subcommand)]
    cmd: Option<Command>
}

#[derive(StructOpt)]
enum Command {
    Bar,
    Baz,
    Quux
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,9 +26,6 @@ pub trait StructOpt {
     /// Returns the corresponding `clap::App`.
     fn clap<'a, 'b>() -> clap::App<'a, 'b>;
 
-    /// Add this app's arguments/subcommands to another `clap::App`.
-    fn augment_clap<'a, 'b>(clap::App<'a, 'b>) -> clap::App<'a, 'b>;
-
     /// Creates the struct from `clap::ArgMatches`.
     fn from_clap(clap::ArgMatches) -> Self;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,8 +10,9 @@
 //! `StructOpt` trait definition
 //!
 //! This crate defines the `StructOpt` trait.  Alone, this crate is of
-//! little interest.  See the `structopt-derive` crate to
-//! automatically generate implementation of this trait.
+//! little interest.  See the 
+//! [`structopt-derive`](https://docs.rs/structopt-derive) crate to
+//! automatically generate an implementation of this trait.
 
 extern crate clap as _clap;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,6 +25,9 @@ pub trait StructOpt {
     /// Returns the corresponding `clap::App`.
     fn clap<'a, 'b>() -> clap::App<'a, 'b>;
 
+    /// Add this app's arguments/subcommands to another `clap::App`.
+    fn augment_clap<'a, 'b>(clap::App<'a, 'b>) -> clap::App<'a, 'b>;
+
     /// Creates the struct from `clap::ArgMatches`.
     fn from_clap(clap::ArgMatches) -> Self;
 
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -228,8 +228,8 @@ fn gen_from_clap(struct_name: &Ident, s: &[Field]) -> quote::Tokens {
     }
 }
 
-fn gen_clap(ast: &DeriveInput) -> quote::Tokens {
-    let struct_attrs: Vec<_> = extract_attrs(&ast.attrs, AttrSource::Struct).collect();
+fn gen_clap(struct_attrs: &[Attribute]) -> quote::Tokens {
+    let struct_attrs: Vec<_> = extract_attrs(struct_attrs, AttrSource::Struct).collect();
     let name = from_attr_or_env(&struct_attrs, "name", "CARGO_PKG_NAME");
     let version = from_attr_or_env(&struct_attrs, "version", "CARGO_PKG_VERSION");
     let author = from_attr_or_env(&struct_attrs, "author", "CARGO_PKG_AUTHORS");
@@ -285,26 +285,41 @@ fn gen_augment_clap(s: &[Field]) -> quote::Tokens {
     }
 }
 
-fn impl_structopt(ast: &syn::DeriveInput) -> quote::Tokens {
+fn impl_structopt_for_struct(name: &Ident, fields: &[Field], attrs: &[Attribute]) -> quote::Tokens {
+    let clap = gen_clap(attrs);
+    let augment_clap = gen_augment_clap(fields);
+    let from_clap = gen_from_clap(name, fields);
+
+    quote! {
+        impl _structopt::StructOpt for #name {
+            #clap
+            #augment_clap
+            #from_clap
+        }
+    }
+}
+
+fn impl_structopt_for_enum(_variants: &[Variant]) -> quote::Tokens {
+    quote!()
+}
+
+fn impl_structopt(ast: &DeriveInput) -> quote::Tokens {
     let struct_name = &ast.ident;
-    let s = match ast.body {
-        Body::Struct(VariantData::Struct(ref s)) => s,
-        _ => panic!("Only struct is supported"),
+    let inner_impl = match ast.body {
+        Body::Struct(VariantData::Struct(ref fields)) =>
+            impl_structopt_for_struct(struct_name, fields, &ast.attrs),
+        Body::Enum(ref variants) =>
+            impl_structopt_for_enum(variants),
+        _ => panic!("structopt only supports non-tuple structs and enums")
     };
 
-    let clap = gen_clap(ast);
-    let augment_clap = gen_augment_clap(s);
-    let from_clap = gen_from_clap(struct_name, s);
     let dummy_const = Ident::new(format!("_IMPL_STRUCTOPT_FOR_{}", struct_name));
     quote! {
         #[allow(non_upper_case_globals, unused_attributes, unused_imports)]
         const #dummy_const: () = {
             extern crate structopt as _structopt;
-            impl _structopt::StructOpt for #struct_name {
-                #clap
-                #augment_clap
-                #from_clap
-            }
+            use structopt::StructOpt;
+            #inner_impl
         };
     }
 }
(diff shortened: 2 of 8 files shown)


williamyaoh:
I'd like some feedback on the documentation, as well as any bugs/kinks that still need to ironed out.

williamyaoh:
I'm not really sure why the Travis build is failing; it seems to still be looking for `structopt-derive 0.0.5` despite having bumped the version(s) in the manifest. `rg '0\.0\.5'` gets nothing, either.

TeXitoi:
To fix Travis, you must update the structopt-derive version in the dev-depedencies of the Cargo.toml at the root of the project.

TeXitoi:
And please update everything to 0.1.0, else we can't do non breaking change version.

williamyaoh:
It's updated at 72d9d35, so it seems to be something else.

TeXitoi:
This commit update only the README.md.

williamyaoh:
Nevermind, you're right, I'm an idiot. I think I accidentally clobbered updating the `[dev-dependencies]` after using `git reset` to fix an older commit.

structopt-derive/src/lib.rs
```diff
@@ -24,28 +24,30 @@
...
+//!   - `about`: Defaults to the crate description given by Cargo.
 //!
-//! Then, each field of the struct correspond to a `clap::Arg`.  As
-//! for the struct attributes, every method of `clap::Arg` in the form
-//! of `fn function_name(self, &str)` can be use in the form of
-//! attributes.  The `name` attribute can be used to customize the
-//! `Arg::with_name()` call (default to the field name).
+//! Then, each field of the struct not marked as a subcommand corresponds
+//! to a `clap::Arg`. As with the struct attributes, every method of
+//! `clap::Arg`in the form of `fn function_name(self, &str)` can be used
```
TeXitoi: Lack space before in

structopt-derive/src/lib.rs
```diff
@@ -74,6 +76,152 @@
...
+//! struct Foo {
+//!   ...
+//!   #[structopt(short = "b")]
+//!   /// The description for the arg that will be displayed when passing `--help`.
+//!   bar: String
+//!   ...
+//! }
+//! ```
+//!
+//! ## Subcomamnds
```
TeXitoi: Typo

structopt-derive/src/lib.rs
```diff
@@ -74,6 +76,152 @@
...
+//!
+//! `structopt` also provides support for applications where certain flags
+//! need to apply to all subcommands, as well as nested subcommands:
+//!
+//! ```ignore
+//! #[derive(StructOpt)]
+//! #[structopt(name = "make-cookie")]
+//! struct MakeCookie {
+//!     #[structopt(name = "supervisor", default_value = "Puck")]
+//!     supervising_faerie: Option<String>,
```
TeXitoi: With default value you don't want an option

ErichDonGubler:
Question about these two lines from the examples:

```rust
struct MakeCookie {
    #[structopt(name = "supervisor", default_value = "Puck")]
    supervising_faerie: Option<String>,
// ...
```

Is it necessary for the `supervising_faerie` field to be an `Option`, if a default is always provided? Wouldn't that just make client code unnecessarily need to call `unwrap` on this field every time?

TeXitoi:
Great doc ! Thanks for the corrections of my frenchy English ;-)

williamyaoh:
@ErichDonGubler You're right, will change that.

ErichDonGubler:
@williamyaoh: Looks like something like this fails right now:

```rust
enum Subcommand {
    Something // Fails here
}
```

Adding curly braces fixes it, but shouldn't be necessary:

```rust
enum Subcommand {
    Something {}
}
```

williamyaoh:
@ErichDonGubler Yup, that's already on the **Things to be improved** list above already. It's fixable, though it's not trivial to do.

williamyaoh:
@ErichDonGubler Fixed at 4ff50e6. It ended up being simpler than I thought it would be.

williamyaoh:
Regarding downcasing of subcommand names by default: I can see the convenience, but on the other hand, it also means using "magic" from the programmer's perspective, where names that they never typed become part of the interface of their program. I'm not sure whether the convenience outweighs the confusion over the exact mechanism of translation of subcommand names.

What does everyone else think?

I think the actual semantics would be to translate all identifiers into kebab case, if no `name` attribute is specified. So

```rust
#[derive(StructOpt)]
enum DoStuff {
    DoTheThing
}
```

would automatically produce a command `$ do-stuff do-the-thing`, without having to specify a `name` attribute. Attributes would act the same way:

```rust
#[derive(StructOpt)]
struct DoOtherStuff {
    useful_information: bool
}
```

would automatically produce a command `$ do-other-stuff --useful-information`.

TeXitoi:
I think translation to kebab-case is an important feature. I prefer this feature to be in another PR, it'll let me the time to validate this one soon.
Maybe open an issue about that to see feedbacks?

src/lib.rs
```diff
@@ -25,6 +26,9 @@ pub trait StructOpt {
     /// Returns the corresponding `clap::App`.
     fn clap<'a, 'b>() -> clap::App<'a, 'b>;
 
+    /// Add this app's arguments/subcommands to another `clap::App`.
+    fn augment_clap<'a, 'b>(clap::App<'a, 'b>) -> clap::App<'a, 'b>;
```
TeXitoi: I'm not a big fan of this, but I don't have any better idea.

williamyaoh: I could move that function into a inherent `impl` block for each type; that would remove it from cluttering up the `StructOpt` documentation, at least, which I think is the real issue.

TeXitoi: Great idea.

williamyaoh:
All right, unless someone else finds a bug in the subcommand-related stuff, that looks to be everything.

williamyaoh:
@TeXitoi Do you need anything else from me before merging?

TeXitoi:
Yes a computer. I'm in vacation.

Le 14 juillet 2017 04:37:20 GMT+02:00, William Yao <notifications@github.com> a écrit :
>@TeXitoi Do you need anything else from me before merging?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/TeXitoi/structopt/pull/17#issuecomment-315253846

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

TeXitoi:
Published, thanks for your work!

PR #22, opened 2017-08-27
Format multiple authors

killercup:
Cargo's `CARGO_PKG_AUTHORS` env var separates author names with a colon, which is not very pretty to look at. So, let's replace it with a comma. Clap has the same functionality it its `crate_authors!` macro but as it's only a few lines of code I didn't want to call that.

I wasn't sure how to add a test for that or if it's needed. I have a project locally where I `[replace]`d the structopt-derive crate with this branch and it works fine.


TeXitoi:
0.1.1 published

killercup:
Ah, no worries, I kinda forgot about it, too :)

PR #26, opened 2017-10-30
Add raw attributes

### LINKED ISSUE #4: Support clap::Arg method not taking just one &str
Several methods are not callable without attr_literals feature gate. And even with.

Flakebi:
This makes it possible to call clap functions that don't take strings,
but any arbitrary value. Even functions that take more than one argument
can be called.

All attributes that are called `attribute_name_raw` are augmented to
`attribute_name(value)` without quoting `value`.

This fixes #4.

For enums, I changed `AppSettings::SubcommandRequired` to `AppSettings::SubcommandRequiredElseHelp` if that's ok?
I find it much more useful if an application lists the help instead of telling me to get the help manually.
Unfortunately this behaviour of structopt cannot be overwritten by the user (no matter which one is used).

### DIFF
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -48,6 +48,8 @@
 //! through specifying it as an attribute.
 //! The `name` attribute can be used to customize the
 //! `Arg::with_name()` call (defaults to the field name).
+//! For functions that do not take a `&str` as argument, the attribute can be
+//! called `function_name_raw`, e. g. `aliases_raw = "&[\"alias\"]"`.
 //!
 //! The type of the field gives the kind of argument:
 //!
@@ -392,7 +394,7 @@ fn gen_augmentation(fields: &[Field], app_var: &Ident) -> quote::Tokens {
             };
             let from_attr = extract_attrs(&field.attrs, AttrSource::Field)
                 .filter(|&(ref i, _)| i.as_ref() != "name")
-                .map(|(i, l)| quote!(.#i(#l)));
+                .map(|(i, l)| gen_attr_call(&i, &l));
             quote!( .arg(_structopt::clap::Arg::with_name(stringify!(#name)) #modifier #(#from_attr)*) )
         });
 
@@ -402,6 +404,20 @@ fn gen_augmentation(fields: &[Field], app_var: &Ident) -> quote::Tokens {
         #( #subcmds )*
         #app_var
     }}
+}
+
+/// Interpret the value of `*_raw` attributes as code and the rest as strings.
+fn gen_attr_call(key: &syn::Ident, val: &syn::Lit) -> quote::Tokens {
+    if let Lit::Str(ref val, _) = *val {
+        let key = key.as_ref();
+        if key.ends_with("_raw") {
+            let key = Ident::from(&key[..(key.len() - 4)]);
+            // Call method without quoting the string
+            let ts = syn::parse_token_trees(val).unwrap();
+            return quote!(.#key(#(#ts)*));
+        }
+    }
+    quote!(.#key(#val))
 }
 
 fn gen_constructor(fields: &[Field]) -> quote::Tokens {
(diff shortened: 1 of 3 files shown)


structopt-derive/src/lib.rs
```diff
@@ -516,19 +544,11 @@ fn gen_augment_clap(fields: &[Field]) -> quote::Tokens {
...
+    let gen = gen_clap(enum_attrs);
     quote! {
         fn clap<'a, 'b>() -> _structopt::clap::App<'a, 'b> {
-            let app = _structopt::clap::App::new(#name)
-                .version(#version)
-                .author(#author)
-                .about(#about)
-                .setting(_structopt::clap::AppSettings::SubcommandRequired);
+            let app = #gen
+                .setting(_structopt::clap::AppSettings::SubcommandRequiredElseHelp);
```
TeXitoi: That's OK for me. @williamyaoh OK?

structopt-derive/src/lib.rs
```diff
@@ -404,6 +406,20 @@ fn gen_augmentation(fields: &[Field], app_var: &Ident) -> quote::Tokens {
...
 }
 
+/// Interpret the value of `*_raw` attributes as code and the rest as strings.
+fn gen_attr_call(key: &syn::Ident, val: &syn::Lit) -> quote::Tokens {
+    if let Lit::Str(ref val, _) = *val {
+        let key = key.as_ref();
+        if key.ends_with("_raw") {
+            let key = Ident::from(&key[..(key.len() - 4)]);
+            // Call method without quoting the string
+            let ts = syn::parse_token_trees(val).unwrap();
```
TeXitoi: I think something like `.expect("bad parameter to {}: the parameter must be valid rust code", key_with_raw)`

structopt-derive/src/lib.rs
```diff
@@ -480,12 +496,28 @@ fn format_author(raw_authors: Lit) -> Lit {
     Lit::Str(authors, StrStyle::Cooked)
 }
 
-fn gen_clap(struct_attrs: &[Attribute], subcmd_required: bool) -> quote::Tokens {
-    let struct_attrs: Vec<_> = extract_attrs(struct_attrs, AttrSource::Struct).collect();
-    let name = from_attr_or_env(&struct_attrs, "name", "CARGO_PKG_NAME");
-    let version = from_attr_or_env(&struct_attrs, "version", "CARGO_PKG_VERSION");
-    let author = format_author(from_attr_or_env(&struct_attrs, "author", "CARGO_PKG_AUTHORS"));
-    let about = from_attr_or_env(&struct_attrs, "about", "CARGO_PKG_DESCRIPTION");
+fn gen_clap(attrs: &[Attribute]) -> quote::Tokens {
```
TeXitoi: 👍 great factoring

structopt-derive/src/lib.rs
```diff
@@ -480,12 +496,28 @@ fn format_author(raw_authors: Lit) -> Lit {
...
+    let attrs: Vec<_> = extract_attrs(attrs, AttrSource::Struct).collect();
+    let name = from_attr_or_env(&attrs, "name", "CARGO_PKG_NAME");
+    let version = from_attr_or_env(&attrs, "version", "CARGO_PKG_VERSION");
+    let author = format_author(from_attr_or_env(&attrs, "author", "CARGO_PKG_AUTHORS"));
+    let about = from_attr_or_env(&attrs, "about", "CARGO_PKG_DESCRIPTION");
+    let settings = attrs.iter()
+        .filter(|&&(ref i, _)| !["name", "version", "auther", "about"].contains(&i.as_ref()))
```
TeXitoi: s/auther/author/

structopt-derive/src/lib.rs
williamyaoh-ah: Seems like it makes `structopt` more usable. No qualms here.

Flakebi:
Thanks for the review, I added some tests and fixed the rest.

TeXitoi:
Thanks! I'll submit to crates.io when I'll have the time. I'll post here. If I didn't do it within 48h feel free to remind me here.

TeXitoi:
0.1.2 published

Issue #39, opened 2017-12-05
Colored help by default ?


bestouff:
Hi,

Would it be possible to have `AppSettings::ColoredHelp` by default, or at least an easy way to enable it without calling into the underlying `clap` ?

TeXitoi:
```rust
extern crate clap;
extern crate structopt;
#[macro_use]
extern crate structopt_derive;

use structopt::StructOpt;

#[derive(StructOpt, Debug)]
#[structopt(setting_raw = "clap::AppSettings::ColoredHelp")]
struct Opt {
    #[structopt(short = "s")]
    speed: bool,
    #[structopt(short = "d")]
    debug: bool,
}

fn main() {
    let opt = Opt::from_args();
    println!("{:?}", opt);
}
```

Does it fix your problem?

bestouff:
Indeed it does, thanks !
I'm sorry but I didn't understood this from the documentation. Maybe an example would help, do you want me to make a patch ?

bestouff:
Update: when trying that:
```rust
  #[derive(StructOpt, Debug)]                                                     
  #[structopt(name = "flashinit", about = "Boot disk populator.",                 
              setting_raw = "clap::AppSettings::ColoredHelp")]                    
  struct Opt {                                                                    
      #[structopt(help = "Disk device or disk image")]                            
      input: String,                                                              
  }                                                                               
```
I've got this:
```rustc
   Compiling flashinit v0.1.0 (file:///home/xav/flashinit)
error[E0433]: failed to resolve. Use of undeclared type or module `clap`
  --> src/main.rs:16:10
   |
16 | #[derive(StructOpt, Debug)]
   |          ^^^^^^^^^ Use of undeclared type or module `clap`

error: aborting due to previous error

error: Could not compile `flashinit`.

To learn more, run the command again with --verbose.
```

TeXitoi:
You have to `extern crate clap;` to expose `clap`.

I'm open to any improvement of the doc, feel free to submit a PR if you are inspired.

I tag this issue to doc, and close it when the doc is improved on this point.

bestouff:
Oh, so I have to also add `clap` as a dependency in `Cargo.toml` and track its version to match the one from structopt to avoid incompatibilities ... that's not an "easy" way anymore.

TeXitoi:
`StructOpt` will always use `clap` version 2. It should be included in clap in the near future.

bestouff:
OK, I've added a note in the doc, see #40

[...]

PR #40, opened 2017-12-05
Add doc on colored help

bestouff:
Added a note about `clap::App`'s _raw methods, and an example specifically about colored help text.


TeXitoi:
Thanks

TeXitoi:
fix #39

PR #43, opened 2018-01-09
Add `bin_name` attribute

sgeisler:
Add support for clap's `.bin_name(&str)` attribute which sets the executable's name shown in the help text. This can be handy for [building cargo subcommands](https://github.com/sgeisler/cargo-remote/blob/master/src/main.rs#L23).

Usage: `#[structopt(bin_name = "my_bin_name")]`

TODO:
- [ ] add documentation
- [ ] add example

### DIFF
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -721,6 +721,8 @@ fn gen_clap(attrs: &[Attribute]) -> quote::Tokens {
     let author = method_if_arg("author", &author);
     let about = from_attr_or_env(&attrs, "about", "CARGO_PKG_DESCRIPTION");
     let about = method_if_arg("about", &about);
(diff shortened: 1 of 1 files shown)


TeXitoi:
Thanks for the contribution!

I think you don't need that, it should work out of the box. Some arguments are handle specifically to get sane default without doing anything, using environment variables. bin_name doesn't need environment variable to have a good default value, and is handled automatically by clap.

If it doesn't work out of the box for you, can you give a example showing what you're trying to fix?

sgeisler:
Oh, thank you very much. I mistook the `settings` at the end for the attribute `settings(Vec<AppSettings>)` from clap. So I didn't test without it.

I'll read the docs again, iirc it's not mentioned that you can use arbitrary attributes. In that case I'd open another pull request to add the documentation.

PR #46, opened 2018-01-23
Allow opting out of clap default features.

ski-csis:
Clap comes with some colorful default features, which are not always useful or wanted, so allow opting out of them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "structopt"
-version = "0.1.6"
+version = "0.1.7"
 authors = ["Guillaume Pinot <texitoi@texitoi.eu>"]
 description = "Parse command line argument by defining a struct."
 documentation = "https://docs.rs/structopt"
@@ -10,11 +10,14 @@ categories = ["command-line-interface"]
 license = "WTFPL"
 readme = "README.md"
 
+[features]
+default = ["clap/default"]
+
 [badges]
 travis-ci = { repository = "TeXitoi/structopt" }
 
 [dependencies]
-clap = "2.20"
+clap = { version = "2.20", default-features = false }
 
 [dev-dependencies]
 structopt-derive = { path = "structopt-derive", version = "0.1.6" }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,15 @@
 //! little interest.  See the 
 //! [`structopt-derive`](https://docs.rs/structopt-derive) crate to
 //! automatically generate an implementation of this trait.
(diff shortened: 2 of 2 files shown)


TeXitoi:
Thanks for the contribution!

I'm not a cargo feature wizard, thus I have a few questions:
 - is the default the same as before? (I suspect yes)
 - How can you cherry-pick clap feature forward structopt? (a note in the documentation about that would be a great addition).

ski-csis:
Yup, this leaves the default unchanged (just checked with a fresh test-crate for good measure :).

I added a note on how to disable the color-stuff in the README, was that where you meant?

TeXitoi:
Mmm, thus, or you desactivate everything, or nothing? You can't desactivate color but activate suggestions?

ski-csis:
I can add features here like:
```
clap-color = ["clap/color"]
clap-suggestions = ["clap/suggestions"]
```
Then you can use `default-features = false, features = ["clap-color"]` to get color w/o suggestions.

But, do you think it's likely anyone would use that?

TeXitoi:
That may be overkill. Let's do it simple first.

README.md
```diff
@@ -85,6 +85,13 @@ $ ./example -ds 1337 foo bar
 Opt { debug: true, speed: 1337, input: "foo", output: Some("bar") }
 ```
 
+If you want to disable the color-terminal support (and other `clap` default-feautres) add `default-features = false` to the `structopt` dependency:
+```toml
+[dependencies]
+structopt = { version = "0.1.0", default-features = false }
+structopt-derive = "0.1.0"
+```
```
TeXitoi: If you want to disable all the `clap` features (color, suggestions...), add [...]

I would prefer this doc to be in src/lib.rs than in README.md.

After that, I'll merge it.

Thanks!

ski-csis:
Like this?

TeXitoi:
Thanks!

TeXitoi:
0.1.7 published

shilangyu:
How do I enable them tho? I read its "on by default" but I don't see colors nor suggestions.

```sh
[dependencies]
structopt = "0.3"
```

CreepySkeleton:
Weird they should be on. 

Try
```
[dependencies]
structopt = { version = "0.3", features = ["color", "suggestions"] }
```

shilangyu:
Nope, nothing. I don't want to spam anymore here, I'll open a separate issue, just last question: Was I supposed to turn on the features explicitly in rust code? Or including them in `Cargo.toml` is supposed to be all?

PR #50, opened 2018-01-29
structopt reexport structopt_derive

### LINKED ISSUE #45: Make structopt-derive's dependency on structopt optional
So, I have a very weird request. I'm trying to re-export stuctopt-derive in a crate, but since this code

https://github.com/TeXitoi/structopt/blob/cc6dfd2e542440e4f194ce7823add4bf09ba9c22/structopt-derive/src/lib.rs#L907-L909

expands to `extern crate structopt`, my consumers would have to add structopt to their dependencies, which I want to avoid. Is there another option of doing this? E.g., making this part of the derive optional (with a feature flag).

cc https://github.com/killercup/quicli/issues/9

TeXitoi:
As using structopt without structopt_derive is quite meaningless, reexport
structopt_derive, symplifying the usage.

Inspired by the failure crate.

fix #45


TeXitoi:
@killercup can you validate that it fixes your case?

killercup:
This looks cool and is probably really nice for people who use structopt directly but since it doesn't change [these lines](https://github.com/TeXitoi/structopt/blob/cc6dfd2e542440e4f194ce7823add4bf09ba9c22/structopt-derive/src/lib.rs#L907-L909) I don't think will affect my case. I'll have a look at it again tomorrow :)

TeXitoi:
These lines are also present (via synstructure) in the failure_derive crate. Thus, if it works for failure, it should work for structopt.

If it works, I suspect you can do the same trick in your crate for serde_derive.

TeXitoi:
I didn't see that you don't really depend on the failure_derive crate for quicli, thus that may not work.

TeXitoi:
If that doesn't fix your issue, I may just remove the extern crate, and only support the `#[macro_use] extern crate structopt;`. But this will be a breaking change, thus it must pass in v0.2 that is in good shape.

structopt-derive/src/lib.rs
```diff
@@ -900,14 +901,5 @@ fn impl_structopt(ast: &DeriveInput) -> quote::Tokens {
         _ => panic!("structopt only supports non-tuple structs and enums")
     };
 
-    let dummy_const = Ident::new(format!("_IMPL_STRUCTOPT_FOR_{}", struct_name));
-    quote! {
-        #[allow(non_upper_case_globals)]
-        #[allow(unused_attributes, unused_imports, unused_variables)]
-        const #dummy_const: () = {
-            extern crate structopt as _structopt;
-            use structopt::StructOpt;
```
TeXitoi: There was a bug here, making the trick quite useless: this line require a module `structopt` at the root that we didn't created in the dummy const. The line should be `use _structopt::StructOpt;` (moreover, this import was useless...).

As using structop-derive without using structopt is quite useless, just remove that as that will fix #45

TeXitoi:
@killercup can you confirm it fix your usecase now?

killercup:
I'm traveling the next few days, will try to get back to you by next Tuesday at the latest! :)

TeXitoi:
OK, thanks and enjoy your trip!

TeXitoi:
@killercup you can test with master, it should do what you need out of the box. master should be the upcoming v0.2.0, thus if you have feedbacks (I think of quicli usage), feel free to tell before the breaking changes are in the wild.

TeXitoi:
@killercup ping. I'll publish v0.2 Saturday, thus I'd like to have your feedback before.

killercup:
Oh! Sorry! I somehow mistook the "merged" state for "feedback no longer necessary"!

TeXitoi:
No problem.

killercup:
So, it seems like I can't re-export the proc-macro from the non-proc-macro structopt crate like I can from the proc-macro structopt_derive crate. (Or, I'm just super confused.)

Also see https://github.com/killercup/quicli/pull/48

PR #56, opened 2018-02-04
Add an u64 value

bestouff:
Prominently add an `u64` value, to show that it works now - without having to RTFM.

(And a typo)


TeXitoi:
You'll also have to update the run example.

TeXitoi:
Maybe just changing speed to u64 would be simpler.

TeXitoi:
I corrected the typo. I'm lazy for adding a u64. Feel Free to reopen a PR if you want.

PR #62, opened 2018-02-10
Add the `from` method to parse from any Iterator

Kerollmops:


src/lib.rs
```diff
@@ -365,4 +367,12 @@ pub trait StructOpt {
     fn from_args() -> Self where Self: Sized {
         Self::from_clap(&Self::clap().get_matches())
     }
+
+    /// Gets the struct from any iterator such as a `Vec` of your making.
+    /// Print the error message and quit the program in case of failure.
+    fn from_any<I, T>(iter: I) -> Self where
+        Self: Sized, I: IntoIterator<Item = T>, T: Into<OsString> + Clone,
```
TeXitoi:
```rust
fn from_iter<I>(iter: I) -> Self
where
    Self: Sized,
    I: IntoIterator,
    I::Item: Into<OsString> + Clone
{
    //...
```

TeXitoi: That was not really about indentation, but about the name of the method and the useless generic.

Kerollmops: Ho ! I will update that, sorry, didn't see.

Kerollmops: done!

TeXitoi:
Thanks for the contribution !

You can update the change log. I'll do it a bit later if you don't have the time.

TeXitoi:
Pups, bad review :-( you didn't change the name to from_iter. I'll modify it later.

PR #64, opened 2018-02-11
Use proper path to 'PathBuf' in README

SergioBenitez:
`PathBuf` is in `std::path`, not `std::fs`.


TeXitoi:
Strange, that should just be a copy and past of the basic example, bit it looks like its unsynchronised.

Thanks for the correction.

PR #77, opened 2018-03-01
Use `proc_macro2::Span::call_site` for all quotes

fitzgen:
This avoids breakage when deriving `StructOpt` when `proc_macro2`'s nightly feature is enabled. See https://github.com/alexcrichton/proc-macro2/issues/67 for details.

cc @alexcrichton

### DIFF
--- a/structopt-derive/src/macros.rs
+++ b/structopt-derive/src/macros.rs
@@ -1,0 +1,11 @@
+// Copyright 2018 Guillaume Pinot (@TeXitoi) <texitoi@texitoi.eu>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+macro_rules! my_quote {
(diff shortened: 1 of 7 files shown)


structopt-derive/src/macros.rs
```diff
@@ -0,0 +1,3 @@
+macro_rules! my_quote {
+    ($($t:tt)*) => (quote_spanned!(::proc_macro2::Span::call_site() => $($t)*))
+}
```
TeXitoi: If you had a file, you should add the header. But maybe for that, just defining it in lib.rs is OK.

TeXitoi:
Travis should also be updated to test with the problematic feature, no?

fitzgen:
Added

* license header
* changelog entry
* travis ci test coverage

alexcrichton:
FWIW This isn't so much a workaround as necessary today based on today's hygiene implementation, otherwise fixing this *without* changing span information would require far more intrusive changes to structopt which'd probably want to be delated for later :)

TeXitoi:
OK. I tried to look at it, but didn't really understand why it doesn't find std. If you know of any doc on that or any similar problem fixed in another project, I would be interested to look at it for my own curiosity.

TeXitoi:
v0.2.5 published

alexcrichton:
@TeXitoi https://github.com/rust-lang/rust/issues/45934 may have a bit more of a writeup if you're interested to dig more!

TeXitoi:
Thanks

PR #83, opened 2018-03-15
Fix raw() sample code

dalance:
The sample code seems to be wrong.
The following error occurred.

```
error[E0433]: failed to resolve. Use of undeclared type or module `structopt`
  --> src/main.rs:16:10
   |
16 | #[derive(StructOpt, Debug)]
   |          ^^^^^^^^^ Use of undeclared type or module `structopt`
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,10 +68,10 @@
 //! #[macro_use]
 //! extern crate structopt;
 //!
-//! use structopt::StructOpt;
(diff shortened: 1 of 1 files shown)


TeXitoi:
Doc comments are already compiled and tested. Thus, the code is working, and this fix is useless.

Can you provide an example reproducing your problem?

dalance:
Sorry, My environment seems to be broken.

PR #87, opened 2018-04-01
Update `syn`, `quote` and `proc-macro2`

hcpl:


TeXitoi:
What's the news about these updates? Can we remove the my_quote macro?

hcpl:
`quote!` 0.5 [now uses `call_site`](https://docs.rs/quote/0.5.1/src/quote/lib.rs.html#204-206), so I'd say yes. I'll try to replace all `my_quote` with `quote`.

PR #88, opened 2018-04-02
Remove meaningless default value on option values

Kerollmops:


Kerollmops:
Tests passes on `rustc 1.26-nightly` on my macbook pro.

TeXitoi:
Please add a line in the changelog

Kerollmops:
We should probably bump the major number because this is not a backward compatible update, it will break previous compiling code. https://semver.org/

TeXitoi:
I've said in 0.2 that I can break compilation for better error handling, thus no major version change for that.

TeXitoi:
The failing test also fail on master. Maybe syn and quote are not yet synchronised with nightly. 

Only the first doc comment is given by syn.

rust-lang/rust#49545 may be related to this.

I didn't understand the problem I do t know where reporting the issue. Maybe syn? @dtolnay any idea?

TeXitoi:
Required is also meaningless with option.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v0.2.7 (unreleased)
+
+* Fail compilation when using `default_value` with `Option` ([#88](https://github.com/TeXitoi/structopt/pull/88)) by [@Kerollmops](https://github.com/Kerollmops)
```
TeXitoi: Not updated accordingly.

dtolnay:
Thanks for the ping. I filed https://github.com/rust-lang/rust/issues/49596 to follow up.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v0.2.7 (unreleased)
+
+* Fail compilation when using `default_value` and `required` with `Option` ([#88](https://github.com/TeXitoi/structopt/pull/88)) by [@Kerollmops](https://github.com/Kerollmops)
```
Kerollmops: `s/and/or` :/

TeXitoi: amended

PR #98, opened 2018-04-20
Add StructOpt::from_iter_safe method

quodlibetor:
I would like to be able to test argument parsing, which requires an error
value, not a `process::exit()`.

### DIFF
--- a/tests/arguments.rs
+++ b/tests/arguments.rs
@@ -10,6 +10,7 @@
 extern crate structopt;
 
 use structopt::StructOpt;
+use structopt::clap;
 
 #[test]
 fn required_argument() {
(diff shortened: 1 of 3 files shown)


tests/arguments.rs
```diff
@@ -8,6 +8,7 @@
 
 #[macro_use]
 extern crate structopt;
+extern crate clap;
```
TeXitoi: Better to use `use structopt::clap;`

quodlibetor:
Done, thanks!

TeXitoi:
Thanks

TeXitoi:
v0.2.8 published

PR #111, opened 2018-05-17
fix Cargo authors list insertion code

windwardly:
Nice crate.
Fix code to transform a list of authors. It was doing no change, leaving colon separated list.
Style it as you like it.
I don't know how to test without also modifying the real authors list in Cargo.toml.
Ideas?


TeXitoi:
The added other to the author list will do the trick.

TeXitoi:
v0.2.9 published.

PR #116, opened 2018-06-06
Add a minimum rustc version policy

### LINKED ISSUE #115: non-reference pattern used to match a reference (see issue #42640)
Hi!

On grabbing 0.2.9 from crates.io last night, I hit the following error several times as it tries to compile structopt-derive:

error[E0658]: non-reference pattern used to match a reference (see issue #42640)
   --> C:\...\structopt-derive-0.2.9\src\attrs.rs:112:17
    |
112 | /                 NameValue(MetaNameValue {
113 | |                     ident,
114 | |                     lit: Str(value),
115 | |                     ..
116 | |                 }) => self.push_str_method(&ident.to_string(), &value.value()),
    | |__________________^
help: consider using a reference

Known issue?

TeXitoi:
Fix #115

@zigguratvertigo @hcpl @killercup OK?

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,12 @@ $ ./basic -o foo.txt -dvvvs 1337 -l alice -l bob --nb-cars 4 bar.txt baz.txt
 Opt { debug: true, verbose: 3, speed: 1337, output: "foo.txt", nb_cars: Some(4), level: ["alice", "bob"], files: ["bar.txt", "baz.txt"] }
 ```
 
+## StructOpt rustc version policy
+
+- Minimum rustc version modification must be specified in the [changelog](https://github.com/TeXitoi/structopt/blob/master/CHANGELOG.md) and in the [travis configuration](https://github.com/TeXitoi/structopt/blob/master/.travis.yml).
+- Contributors can increment minimum rustc version without any justification if the new version is required by the latest version of one of StructOpt's depedencies (`cargo update` will not fail on StructOpt).
+- Contributors can increment minimum rustc version if the library user experience is improved.
+
 ## Why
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -102,6 +102,12 @@ $ ./basic -o foo.txt -dvvvs 1337 -l alice -l bob --nb-cars 4 bar.txt baz.txt
 Opt { debug: true, verbose: 3, speed: 1337, output: "foo.txt", nb_cars: Some(4), level: ["alice", "bob"], files: ["bar.txt", "baz.txt"] }
 ```
 
+## StructOpt rustc version policy
+
+- Minimum rustc version modification must be specified in the [changelog](https://github.com/TeXitoi/structopt/blob/master/CHANGELOG.md) and in the [travis configuration](https://github.com/TeXitoi/structopt/blob/master/.travis.yml).
+- Contributors can increment minimum rustc version without any justification if the new version is required by the latest version of one of StructOpt's depedencies (`cargo update` will not ffail on StructOpt).
```
killercup: s/ffail/fail

README.md
```diff
@@ -102,6 +102,12 @@ $ ./basic -o foo.txt -dvvvs 1337 -l alice -l bob --nb-cars 4 bar.txt baz.txt
...
+- Contributors can increment minimum rustc version if the library user experience is improved.
+
```
killercup: So it's cool to bump the min required rustc version between structopt 1.2.3 to 1.2.4? I'd probably try and do that in minor versions only. This is also the policy in some other crates

TeXitoi: Right now, we are in 0.x version, thus I don't want to do a breaking change version to upgrade rustc.

But, after 1.0.0 I think that following the minor rule would bethe way to go.

TeXitoi:
v0.2.10 published.

PR #118, opened 2018-06-17
Explicitly mention doc comments in flatten error

### LINKED ISSUE #117: Unclear error message when flatten is used with doc comment
Something like the following (from [quicli#84](https://github.com/killercup/quicli/issues/84)):
```
#[derive(Debug, StructOpt)]
struct Cli {
    /// Increase verbosity
    #[structopt(flatten)]
    verbosity: Verbosity
}
```

Will give the error:
```
methods are not allowed for flattened entry
```

It would be very helpful if the error message was explicit about the doc comments possibly being the problem as well. Thanks.

spease:
Fix #117


TeXitoi:
Thanks.

PR #120, opened 2018-06-27
Simplify `group` example

bruceadams:
This does change the behavior of the `group` example. The `vers` group is no longer required. It is is still a group, that is: only one of the several options are accepted.

I'm not sure this is an improvement to the example. It is much simpler and aligns with the ways I usually use groups, where they are not required.

### DIFF
--- a/examples/simple_group.rs
+++ b/examples/simple_group.rs
@@ -1,0 +1,31 @@
+#[macro_use]
+extern crate structopt;
+
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug)]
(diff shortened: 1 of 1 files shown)


TeXitoi:
I don't want to remove this example because it explains how to do the difficult thing. But I'll accept a new example simple_group.rs with that.

bruceadams:
Makes sense. The existing example does show interesting stuff. I redid this pull request to add a simple group example.

TeXitoi:
Thanks for the contribution!

PR #140, opened 2018-10-05
Upgrade syn to 0.15

konstin:
This removes the dependency on syn 0.14 in downstream creates, which means you don't need to compile syn 0.14 and 0.15. Surprisingly, this update worked by just increasing the version number.


TeXitoi:
Thanks

TeXitoi:
v0.2.11 published

PR #144, opened 2018-10-13
Fix use of validators

rylev:
It's seems the use of validators isn't really supported at all.

With the following example:
```rust
#[derive(StructOpt, Debug)]
struct Opt {
    #[structopt(validator = my_validator)]
    foo: usize
}

fn my_validator(x: String) -> Result<(), String> {
    Ok(())
}
```

I get the following errror:
```
  --> src/main.rs:20:10
   |
20 | #[derive(StructOpt, Debug)]
   |          ^^^^^^^^^
   |
   = help: message: invalid structopt syntax: attr
```

This PR fixes both the error case of trying to use an identifier in the value position, and actually supporting the validator use case.

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -76,6 +76,13 @@ impl Attrs {
                 self.methods = methods.into_iter().filter(|m| m.name != name).collect();
             }
             ("name", new_name) => self.name = new_name.into(),
+            ("validator", _) => {
+                use quote::ToTokens;
+                self.methods.push(Method {
+                    name: name.to_string(),
+                    args: Ident::new(arg, Span::call_site()).into_token_stream(),
+                });
+            },
             (name, arg) => self.methods.push(Method {
                 name: name.to_string(),
                 args: quote!(#arg),
@@ -86,16 +93,19 @@ impl Attrs {
         use Lit::*;
         use Meta::*;
         use NestedMeta::*;
+        let name = self.name.clone();
 
         let iter = attrs
             .iter()
             .filter_map(|attr| {
                 let path = &attr.path;
                 match quote!(#path).to_string() == "structopt" {
-                    true => Some(
-                        attr.interpret_meta()
-                            .expect(&format!("invalid structopt syntax: {}", quote!(attr))),
-                    ),
+                    true => {
+                        let meta = attr
+                            .interpret_meta()
+                            .expect(&format!("invalid structopt attribute syntax for `{}`", name));
+                        Some(meta)
+                    },
                     false => None,
                 }
             }).flat_map(|m| match m {


TeXitoi:
I'll review your PR, but validator is handled automatically by structopt. If you need custom validator, look at try_from attribute.

structopt-derive/src/attrs.rs
```diff
@@ -76,6 +76,13 @@ impl Attrs {
                 self.methods = methods.into_iter().filter(|m| m.name != name).collect();
             }
             ("name", new_name) => self.name = new_name.into(),
+            ("validator", _) => {
```
TeXitoi: I don't want special casing for normal method.

structopt-derive/src/attrs.rs
```diff
@@ -76,6 +76,13 @@ impl Attrs {
...
+                use quote::ToTokens;
+                self.methods.push(Method {
+                    name: name.to_string(),
+                    args: Ident::new(arg, Span::call_site()).into_token_stream(),
```
TeXitoi: An identity is very llimiting. If custom attributes now support identity, it must work for any command.

rylev:
@TeXitoi " validator is handled automatically by structopt" doesn't seem to be 100% true. You can mimic the same functionality with a custom parser, but I cannot find a way to just use a validator. The docs say: "So derive(StructOpt) tells Rust to generate a command line parser, and the various structopt attributes are simply used for additional parameters.", and this breaks if you try to use "validator" as an attribute.

structopt-derive/src/attrs.rs
rylev: I'm not sure what you mean by "custom attributes" - can you explain more?

structopt-derive/src/attrs.rs
rylev: This is where we map from key/value attribute pairs (where the value is a string) to methods. If we don't special case here, we'd have to special case somewhere else. The `arg` to  the `validator` method is an identifier (or a closure, but that should be handled somewhere else). 

In other words, I'm not sure how else you would expect to handle this.

rylev:
@TeXitoi we unfortunately cannot make structopt work when an identifier is in the value position of an attribute, because that is a syntax error. Rust does not allow it. Unfortunately there's nothing we can about that: https://doc.rust-lang.org/reference/attributes.html

rylev:
@TeXitoi Yea that works for sure. Do you have any interest in getting the more intuitive version of validator = "my_validator" working?

TeXitoi:
I don't wwant to special case validator as else, we should special case a lot of method, and it will them be a nightmare tto maintain, document and use.

rylev:
@TeXitoi Ok it's your call. I think keeping the improved error message is worth it - I can also change the docs to make this use case a bit clearer. Does that sound good?

TeXitoi:
OK for improving error message and doc. Doc is always great!

Thanks.

Issue #134, opened 2018-08-26
Feature request: smarter parsing of doc comments


Lucretiel:
If I have a docopt like this:

```rust
#[structopt(short="n", long="no-op")]
/// Run the command as normal, but don't modify the file
///
/// In no-op mode, rewrite will do everything it normally does, including
/// writing to a temporary file, if enabled. The only thing it doesn't do is
/// modify the contents of the file.
no_op: bool,
```

It would be nice if structopt detected the line break, and set the separate parts as [`help`](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.help) and [`long_help`](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.long_help). It would also be nice if it automatically reflows the line breaks correctly.

TeXitoi:
Good idea, but it will be a breaking change.

Structopt already merge lines, thus, clap should reflow. If not, that a problem with clap.

0ndorio:
Would it be possible to come up a new feature flag (e.g. `split_doc_comment`, `upcoming_enhancement`) to introduce such a new feature earlier as an explicit opt-in?

TeXitoi:
That's not a problem of doing a breaking change, we just release a 0.3 and that's it. It will force me to do the other breaking changes that I want (I think of #18).

0ndorio:
> Structopt already merge lines, thus, clap should reflow. If not, that a problem with clap.

Maybe somthing completly differnt but as it fits the `smarter doc comment parsing` section. Is there any option forcing structopt to keep the formatting of the doc comment to achieve the same behaviour as using a raw string for `long_help` or `long_about`?

This could be nice as it would allow to introduce proper formatted doc comments including lists and source blocks into the long descriptions.

ebkalderon:
Since #146 was already merged and published, shouldn't this issue be closed?

TeXitoi:
@ebkalderon good point

@0ndorio if you still want https://github.com/TeXitoi/structopt/issues/134#issuecomment-429464231 please create a dedicated issue.

Lucretiel:
Thanks!

PR #146, opened 2018-10-26
Introduce smarter parsing of doc comments.

0ndorio:
First approach of an implementation of the smarter doc comment parsing feature requested in #134.

It introduces an additional check in the doc comment parsing algorithm to test if the second line of a given doc comment is an empty line. If this is true only the first line will be used as input for the argument given as input parameter (`help` or `about`) at the same time it applies the whole comment as input for the related long version (`long_help` and `long_about`).

This patch *doesn't* contain the additional feature I requested in [this comment](https://github.com/TeXitoi/structopt/issues/134#issuecomment-429464231). It would be easy to add it but I assume as long as it isn't clearly defined how to switch between the two behaviors it should not be introduced.

### DIFF
--- a/examples/doc_comments.rs
+++ b/examples/doc_comments.rs
@@ -1,0 +1,86 @@
+// Copyright 2018 Guillaume Pinot (@TeXitoi) <texitoi@texitoi.eu>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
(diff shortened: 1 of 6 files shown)


0ndorio:
I can try to add the missing docs. Which level of documentation do you miss? User facing or inside the updated parsing logic?

TeXitoi:
For the user, something like if there is a line followed by an empty line, the first will be help, the rest long_help

0ndorio:
> For the user, something like if there is a line followed by an empty line, the first will be help, the rest long_help

Added something. Feel free to nitpick.

0ndorio:
@TeXitoi ping :)

TeXitoi:
v0.2.14 published

PR #148, opened 2018-11-10
Fix up some docs about argument renaming.

adeschamps:
The only error was that the docs claimed the default style for long arguments was kebab-case, when it's actually literal. The rest is just verbage and punctuation.


TeXitoi:
Thanks !

PR #154, opened 2018-12-08
Fix invalid structopt syntax error message

roblabla:
There was a typo preventing the probematic attr to be shown to the user.


TeXitoi:
Thanks. Too bad I didn't see it this morning. Will be in the next release.

PR #161, opened 2019-01-09
Remove line-ending periods from short help.

TedDriggs:
This makes the generated short help more natural for the command-line,
while allowing the rustdoc text to have the period that's standard for
rustdoc comments.


TeXitoi:
Thanks. Tell me if you want a release, else it'll wait for the next one.

PR #169, opened 2019-03-11
add completions generation example

vn971:


vn971:
This PR comes out of my personal struggling: I've spent quite some time figuring out how to generate completions, all until I've found the `clap` method in source code which I was reading from the cloned repo by that time. It could be easy for some people maybe, but it was hard to find for me.

TeXitoi:
Thanks!

PR #179, opened 2019-04-12
Suppress `clippy::redundant_closure`

### LINKED ISSUE #177: clippy warning: redundant closure found
```rust
/// Manylinux2
pub enum Manylinux2 {}

impl std::str::FromStr for Manylinux2 {
    type Err = &'static str;

    fn from_str(_: &str) -> Result<Self, Self::Err> {
        unreachable!()
    }
}

#[derive(StructOpt)]
pub struct BuildOptions2 {
    pub manylinux: Manylinux2,
}
```

This results in the following clippy warning:

```
warning: redundant closure found
  --> src/build_options.rs:32:10
   |
32 | #[derive(StructOpt)]
   |          ^^^^^^^^^
   |
   = note: #[warn(clippy::redundant_closure)] on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
```

qryxip:
Fixes #177. We cannot change `.map_err(|e| e.to_string())` to `.map_err(ToString::to_string)` since `e` is not a reference in most cases. Instead, we have to wirte

- `.map_err(|e| (&e).to_string())`
or
- `.map_err(|e| ::std::string::ToString::to_string(&e))`

to suppress the warning.

The upstream clippy has already stopped triggering the warning for macros. This hack will be obsolete, but the current stable clippy will keep producing the warning until Rust 1.35 is released.
rust-lang/rust-clippy/pull/3816

### DIFF
--- a/examples/enum_in_args.rs
+++ b/examples/enum_in_args.rs
@@ -1,8 +1,10 @@
-#[macro_use]
-extern crate structopt;
-#[macro_use]
-extern crate clap;
+//! current `clap::arg_enum!` uses "non-ident macro path" feature, which was stabilized in
+//! Rust 1.31.0.
 
+extern crate clap;
+extern crate structopt;
+
+use clap::arg_enum;
 use structopt::StructOpt;
 
 arg_enum! {
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
 language: rust
 cache: cargo
-rust:
-  - 1.21.0
-  - stable
-  - beta
-  - nightly
 matrix:
   include:
+    - rust: 1.21.0
+      # `examples/enum_in_args.rs` cannot be compiled because of `clap::arg_enum!`.
+      env: TARGETS=--tests
+    - rust: stable
+    - rust: beta
+    - rust: nightly
     - rust: nightly
       env: FEATURES="--features nightly"
     - rust: stable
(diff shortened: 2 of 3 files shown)


examples/enum_in_args.rs
```diff
@@ -1,16 +1,57 @@
...
+// ```
+// clap::arg_enum! {
+//     #[derive(Debug)]
+//     enum Baz {
+//         Foo,
+//         Bar,
+//         FooBar,
+//     }
+// }
+//```
```
TeXitoi: I don't want to remove this example, a lot of person look at it. If that's a problem about the minimum rust version, as clap has updated its version, we can as described in the policy.

qryxip: <https://travis-ci.org/TeXitoi/structopt/jobs/519234259#L465>
I had thought `tests` compiled and executed `examples`. We don't need to modify `enum_in_args` because we can avoid the Travis error by adding just `--tests` option for `1.21.0` to skip compiling `examples/`. Though [`clap` has already stopped supporting `1.21.0`](https://github.com/clap-rs/clap/pull/1397), I cannot think of any other reason to bump the Rust version. `structopt` and its tests work for now.

TeXitoi:
Thanks.

konstin:
Thank you!

PR #181, opened 2019-04-24
Fix use of deprecated string functions

owenthewizard:


TeXitoi:
For that, the minimum rust version must be changed.

owenthewizard:
Travis Rust is too old?

owenthewizard:
Sorry, didn't even see your comment. When/if do you anticipate a minimum version bump?

TeXitoi:
As of the structopt policy, we can update to the supported clap version when we want. See the readme.

owenthewizard:
I don't think this "improves the library experience" per se, so perhaps merging this can wait until something more worthwhile bumps the version.

TeXitoi:
I've cherry-picked the interesting commit

PR #189, opened 2019-05-28
Update deprecated trim_{left,right}_matches functions

sphynx:
Those functions were deprecated in Rust 1.33.0, so I just decided to get rid of the warning messages.


sphynx:
Oops, someone already fixed that ;)

PR #190, opened 2019-05-28
Support Option<Option<T>> field types

### LINKED ISSUE #188: Support optional arguments for options
Sometimes options have an optional argument, for example in `rustfmt` we can use either `--help` or `--help=<topic>`.

In order to achieve this we can use `Arg::min_values(0)` in `clap` (as described here: https://github.com/clap-rs/clap/issues/892), but there is no way to directly specify it in `structopt`. Using `Option<String>` type for a field would fail to distinguish between no option provided at all and an option without an argument. It will return `None` for both. 

It looks like a possible way to support it would be to use `Option<Option<String>>` field type which can lead to `min_values(0)` in generated `clap` code.

Note that this type of options is also supported by `getopts` library:
https://doc.rust-lang.org/1.1.0/getopts/fn.optflagopt.html

This issue seems to be related to #180, but it has `Option<Vec<T>>` instead of `Option<Option<T>>`

sphynx:
Closes #188 

I used `sub_type` function in `Attrs` module, so I moved it to there and exported so that it can be reused in `lib.rs`, I am not sure this is an optimal solution, please let me know if you want to change that (for example to have a separate module for types manipulations or anything else).

Another thing that I wanted to add to tests: is to make sure that `Option<Option<T>>` cannot be used for positional arguments. In this case I would expect a macro code-generation to fail, but I'm not sure how to make a test which expects particular macro failure. For now, I just checked that it works manually.

Also please note that I am using `max_values=1` as well as `min_values=0` to limit number of values to 1 in the generated `clap` code. However, handling of `max_option` in `clap` apparently has a bug: it tries to consume one more argument than it should during the parsing. 

I fixed that bug and prepared a PR here: https://github.com/clap-rs/clap/pull/1481  
However, before that bug is fixed, theoretically we can have problems if, say, `Option<Option<T>>` long field with a value is immediately followed by a subcommand, like this:

```
test --field 1 add -p 2
```

then `clap` (without my fix) would continue consuming arguments for `--field` even when `max_values=1` is set, so it will consume subcommand `add` as well and parsing will fail on seeing `-p`. But this is not directly related to `Option<Option<T>>` feature, just something to keep mind since it's related to `max_values`.

Please let me know if you have any suggestions about the code.

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -404,12 +404,9 @@ impl Attrs {
         {
             match segments.iter().last().unwrap().ident.to_string().as_str() {
                 "bool" => Ty::Bool,
-                "Option" => match sub_type(ty) {
-                    Some(t) => match Attrs::ty_from_field(t) {
-                        Ty::Option => Ty::OptionOption,
-                        _ => Ty::Option,
-                    },
-                    None => Ty::Option,
+                "Option" => match sub_type(ty).map(Attrs::ty_from_field) {
+                    Some(Ty::Option) => Ty::OptionOption,
+                    _ => Ty::Option,
                 },
                 "Vec" => Ty::Vec,
                 _ => Ty::Other,
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -10,7 +10,10 @@ use heck::{CamelCase, KebabCase, MixedCase, ShoutySnakeCase, SnakeCase};
 use proc_macro2::{Span, TokenStream};
 use std::{env, mem};
 use syn::Type::Path;
-use syn::{self, Attribute, Ident, LitStr, MetaList, MetaNameValue, TypePath};
+use syn::{
+    self, AngleBracketedGenericArguments, Attribute, GenericArgument, Ident, LitStr, MetaList,
+    MetaNameValue, PathArguments, PathSegment, TypePath,
+};
 
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub enum Kind {
@@ -23,6 +26,7 @@ pub enum Ty {
     Bool,
     Vec,
     Option,
+    OptionOption,
     Other,
 }
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


TeXitoi:
you need to rustfmt please.

structopt-derive/src/attrs.rs
```diff
@@ -400,7 +404,13 @@ impl Attrs {
...
             match segments.iter().last().unwrap().ident.to_string().as_str() {
                 "bool" => Ty::Bool,
-                "Option" => Ty::Option,
+                "Option" => match sub_type(ty) {
+                    Some(t) => match Attrs::ty_from_field(t) {
+                        Ty::Option => Ty::OptionOption,
+                        _ => Ty::Option,
+                    },
+                    None => Ty::Option,
+                },
```
TeXitoi:
```rust
"Option" => match sub_type(ty).map(Attrs::ty_from_field) {
    Some(Ty::Option) => Ty::OptionOption,
    _ => Ty::Option,
},

TeXitoi:
No problem for the `sub_type` move.

There is no (yet) compile failure tests, so I don't ask to do this. But I'll appreciate a dedicated PR adding this new infrastructure ;-)

Noted for the clap problem, can be useful if a related bug report rises.

TeXitoi:
Also can you update the CHANGELOG.md? Thanks.

TeXitoi:
I forgot a last thing: this feature must be documented.

structopt-derive/src/attrs.rs
sphynx: Sure.

sphynx:
I've fixed everything you mentioned. 

Not sure about CHANGELOG format: should it be under a new version header? For now I've just added a new item on the very top without the version header. I can bump the package version as well (but perhaps you'd like to merge other pull requests first).

Thanks for your feedback!

TeXitoi:
Thanks!

TeXitoi:
v0.2.16 published

Issue #187, opened 2019-05-26
structopt as main args through rust-cli/paw


yoshuawuyts:
A few weeks ago the CLI WG released [paw](https://github.com/rust-cli/paw), a crate to try and make CLI parsing more of a first-class citizen in Rust ([post](https://blog.yoshuawuyts.com/paw/)). This allows arguments to be passed into `fn main` as long as a trait is implemented. Obviously we also added [structopt support](https://github.com/rust-cli/paw/tree/master/paw-structopt).

So far it's been working quite well, but we think we could make this even better if we could add the trait to structopt directly:

__Proposed__
```rust
#[derive(structopt::StructOpt)]
struct Args {}

#[paw::main]
fn main(args: Args) {
    println!("{}", args);
}
```

__Current__
```rust
#[derive(paw_structopt::StructOpt, structopt::StructOpt)]
struct Args {}

#[paw::main]
fn main(args: Args) {
    println!("{}", args);
}
```
## Implementation
The patch to `structopt` would be about 6 lines inside the structopt macro:
```rust
impl paw::ParseArgs for #name {
    type Error = std::io::Error;
    fn parse_args() -> Result<Self, Self::Error> {
        Ok(<#name as structopt::StructOpt>::from_args())
    }
}
```

Would you be open to a patch that allows structopt to be used directly from main?

---

## Full TCP Example
```rust
#[derive(structopt::StructOpt)]
struct Args {
    #[structopt(flatten)]
    port: clap_port_flag::Address,
    #[structopt(flatten)]
    port: clap_port_flag::Port,
}

#[paw::main]
fn main(args: Args) -> Result<(), std::io::Error> {
    let listener = TcpListener::bind((&*args.address, args.port))?;
    println!("listening on {}", listener.local_addr()?);
    for stream in listener.incoming() {
        stream?.write(b"hello world!")?;
    }
    Ok(())
}
```

TeXitoi:
Can't you just `impl<T> ParseArg for T where T: StructOpt`

yoshuawuyts:
@TeXitoi unfortunately we can't because we're not allowed to implement traits for foreign types (believe it's because of orphan rules):

![2019-05-27-114218_1920x1080](https://user-images.githubusercontent.com/2467194/58411380-77718200-8074-11e9-8337-2f8425c1cdd7.png)

Maybe I'm missing something tho?

TeXitoi:
[...]

PR #192, opened 2019-06-01
Optionally add paw support for structopt as a feature

gameldar:
Create an optional feature that when turned on add in the required paw:ParseArgs trait implementation that allows use of StructOpt as the argument for the fn main.


gameldar:
Note: this resolves  #187

TeXitoi:
A generic impl in `src/lib.rs` doesn't work?

gameldar:
> A generic impl in `src/lib.rs` doesn't work?
Oh that was a quick reply! I just wrote that on the bug  - but no - you end up with the same error  @yoshuawuyts saw that you can't use generic type T on something from a different crate.

TeXitoi:
OK.

Please update the CHANGELOG, and I'll merge. A new version should be published within a week, waiting for a doc improvement.

TeXitoi:
thanks!

TeXitoi:
v0.2.17 published

PR #195, opened 2019-06-01
Docs: tell about possible literals use for attribute values

sphynx:
For some reason, the possibility of using numbers and booleans in attribute values has not been mentioned in the documentation. Perhaps, it was not supported before or doesn't work with some older Rust versions. 

I've added it to the documentation, but please let me know if there any hidden issues here.


TeXitoi:
At the time of the feature creation, only strings was allowed. Using bool and usize worked without modification with a new rustc that support bool and usize as attributes.

TeXitoi:
v0.2.17 published

PR #196, opened 2019-06-03
add default_value to arg_enum example

vn971:


examples/enum_in_args.rs
```diff
@@ -19,7 +19,11 @@ arg_enum! {
 #[derive(StructOpt, Debug)]
 struct Opt {
     /// Important argument.
-    #[structopt(raw(possible_values = "&Baz::variants()", case_insensitive = "true"))]
+    #[structopt(raw(
+        possible_values = "&Baz::variants()",
+        case_insensitive = "true",
+        default_value = "\"FooBar\"",
+    ))]
```
TeXitoi: It complexify the example.

IMHO, the correct way to do that would be (today, not supported by the minimal rust version)
```rust
 #[derive(StructOpt, Debug)] 
struct Opt {
    #[structopt(
        raw(possible_values = "&Baz::variants()"),
        case_insensitive = true,
        default_value = "FooBar"
    )]
    i: Baz,
}
```

TeXitoi:
Thanks for the interest and the proposition, but I'm not sure that's the good way of improving this example. What was your misunderstanding that conduct you to propose this change?

examples/enum_in_args.rs
vn971: Yes, that's exactly the point. You _would want_ the simpler version to work, so you type in `default_value = "FooBar"`. This is the compilation error that you get now:

```
error[E0425]: cannot find value `FooBar` in this scope
  --> src/cli_args.rs:14:10
   |
14 | #[derive(StructOpt, Debug)]
   |          ^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: Could not compile `rua`.
```

What does it mean? How to fix it? The fix was not immediately obvious to me and only came to me later, accidentally. By having this documented I hope to help people fight with the potential problem. Be aware of the problem and the fix rather than hiding both.

TeXitoi:
OK, but this solution only solve the case of adding a structopt attribute using an enum, not mixing raw and non raw attributes. Can you think of a more general solution that would help to find the fix? I'd prefer improvement of the documentation that tweeking one of the examples.

examples/enum_in_args.rs
TeXitoi: And the version I gave is working, you just need a recent rust version.

vn971: @TeXitoi ah, indeed, I just missed the braces placement for `raw()`!

Do you know if it's possible to change the old form of
```
#[structopt(raw(possible_values = "&Baz::variants()", case_insensitive = "true"))]
```
to
```
#[structopt(raw(possible_values = "&Baz::variants()"), case_insensitive = true)]
```
in the old compiler then? This way, just appending `, default_value = "whatever"` at the end would not break anything.

Or does it feel more like an edge case, and many other developers clearly understand the difference between inside `raw` and outside of it?

TeXitoi: I think that's not possible with rust 1.21 which is the current oldest supported version by structopt. But this version might change within a few month.

vn971: Thanks for all explanations!

Last question, shouldn't the import be `use structopt::clap::arg_enum` instead of `use clap::arg_enum`?

TeXitoi: I think it doesn't work on older version because of macro reexport limitation, but that's better to do so if possible.

PR #198, opened 2019-06-08
Custom attributes parser

### LINKED ISSUE #178: Attribute token streams for raw calls (Rust 1.34)
Hey, I've just read about token streams being enabled in custom attributes: https://blog.rust-lang.org/2019/04/11/Rust-1.34.0.html#custom-attributes-accept-arbitrary-token-streams

Might be useful for the raw clap calls like `raw(global_settings = "&[AppSettings::DisableVersion]")`

Probably doesn't warrant the version bump I guess, but would be nice to keep in mind for the future.

sphynx:
Closes #178 

I've added a new module: `parse.rs`, which contains an intermediate representation for `structopt` attributes. This representation reflects domain-specific attributes of `structopt` and is used instead of generic `syn.Meta`, which played a similar role in the previous version of the code. This allows parsing arbitrary expressions instead of String literals in attribute values. I tried to keep those intermediate structures low-level, i.e. closer to the token level, so that it's easy to write `Parse` trait implementations for them. Later, they are converted to more high-level representation in `attrs` module.

I didn't change the parsing of doc-comments, since they are not affected by the introduction of arbitrary tokens in attributes in Rust 1.34. So `syn.Meta` is still used there.

Please take a look at the code: if you like the approach, I'll add some notes to the documentation and the ChangeLog.

Thank you!


sphynx:
My newly added tests obviously can't be compiled on Rust 1.21.0, because the feature in question is supported in 1.34.0.  I'm not sure how to proceed with those tests in the best way. 

The issue is similar to `arg_enum!` from `clap` which is present in `examples`., but it's a test, not an example. I'm not sure how to disable a single test module (can't find anything useful in `cargo test --help`). 

Perhaps I can `[ignore]` those tests by default and turn them on in all the cases except Rust 1.21.0? It's not ideal though.  Or I can try to introduce a custom build attribute and use it with `cfg`?

TeXitoi:
We can just skip the tests on rust 1.21. If it compils, and tests are passing using latest stable rust, there is no reason that it would not work with 1.21 for the usable features.

TeXitoi:
I may take a bit of time to review this PR as it's quite long.

sphynx:
> We can just skip the tests on rust 1.21. If it compils, and tests are passing using latest stable rust, there is no reason that it would not work with 1.21 for the usable features.

Ok, so we should update `.travis.yml` for that, right? I'm not sure how to do it, my guess is that overriding `script` for 1.21.0 can work:

```
    - rust: 1.21.0
      # just build the code for 1.21.0, since some of the tests use features
      # which are unsupported in 1.21.0
      script: cargo build
```

Is it the right way to do it in Travis? If so, I can add that change to this PR.

sphynx:
> Is it the right way to do it in Travis? If so, I can add that change to this PR.

Ok, I've changed `.travis.yaml`, we'll see how it goes :)

TeXitoi:
Yep, it's working.

sphynx:
Note that 'full' version of `syn` is needed for this change to support parsing of more complicated expressions like `&[AppSettings::Foo, AppSetting::Bar]`. 

A quote from `syn` documentation:

> `full` — Data structures for representing the syntax tree of all valid Rust source code, including items and expressions.

TeXitoi:
As noted in #200, v0.2.17 was a breaking change version. Thus, I've yanked the version, and updated the minimal rust version to 1.34. The plan is to put the new version in v0.3. v0.3 "need" a edition 2018 before publishing.

structopt-derive/src/parse.rs
```diff
@@ -0,0 +1,158 @@
...
+pub struct StructOptAttributes {
+    pub paren_token: syn::token::Paren,
+    pub attrs: Punctuated<StructOptAttr, Token![,]>,
+}
+
+impl Parse for StructOptAttributes {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let content;
+        Ok(StructOptAttributes {
+            paren_token: parenthesized!(content in input),
```
TeXitoi: From where is this macro? syn?

tests/custom-string-parsers.rs
```diff
@@ -217,7 +217,7 @@ fn test_custom_bool() {
             parse(try_from_str = "parse_bool")
         )]
         verbose: bool,
-        #[structopt(short = "t", parse(try_from_str = "parse_bool"))]
+        #[structopt(short = "t", parse(try_from_str = parse_bool))]
         tribool: Option<bool>,
         #[structopt(short = "b", parse(try_from_str = "parse_bool"))]
```
TeXitoi: As we are in a breaking change version, we can only support the quote less version

tests/non_literal_attributes.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    #[structopt(short = "l", long = "level", aliases = &["set-level", "lvl"])]
+    level: String,
+
+    #[structopt(long = "values")]
+    values: Vec<i32>,
+
+    // Note that `raw(...)` is still needed here, for non-convential
+    // methods taking more than one argument.
+    #[structopt(name = "FILE", raw(requires_if = r#""FILE", "values""#))]
```
TeXitoi: It would be great to support this use case, and remove the raw attribute. As we are on a breaking change version, we can change the syntax. Any idea?

structopt-derive/src/parse.rs
sphynx: Yes, it's from `syn`: https://docs.rs/syn/0.15.35/syn/macro.parenthesized.html

tests/custom-string-parsers.rs
sphynx: OK, makes sense.

tests/non_literal_attributes.rs
sphynx: One idea is to allow tuples as attributes values and automatically transform them into multiple arguments. As far as I can see, no Clap functions accept tuples, so that should not lead to any conflicts, but theoretically can break later, if Clap decides to introduce a method with a tuple argument on `Arg` or `App`. But that is unlikely.

I mean something like this:

```
#[structopt(name = "FILE", requires_if = ("FILE", "values"))]
```

TeXitoi: what about supporting `method(arg1, arg2)` with `method = arg1` only a shortcut for `method(arg1)`?

That would be:
```rust
#[structopt(name = "FILE", requires_if("FILE", "values"))]

Comment (sphynx) [reply tests/non_literal_attributes.rs]: Yes, I think your approach is better.

Comment (sphynx) [reply tests/non_literal_attributes.rs]: I've implemented your approach and removed any mentions of raw attributes, since they are not needed anymore.

File: tests/custom-string-parsers.rs (hunk 2 above)
Comment (sphynx) [reply tests/custom-string-parsers.rs]: Done

**Comment by sphynx** [comment]
I've implemented the requested changes, could you please take a look again?

Thank you!

**Comment by sphynx** [comment]
I've improved the error messages from `parse.rs`, deleted obsoleted `trybuild` test cases and rebased several others to contain updated error messages.

As a side note: this pull request is getting longer and longer (already at 18 commits!) and it's becoming harder to maintain: to rebase it or to merge the other changes in...

**Comment by TeXitoi** [comment]
Yes, sorry, that's the priority now. Will try to review it soon. No other PR will be merged before this one.

File: tests/arguments.rs [hunk 4]
```diff
@@ -56,7 +56,7 @@ fn argument_with_default() {
 fn argument_with_raw_default() {
     #[derive(StructOpt, PartialEq, Debug)]
     struct Opt {
-        #[structopt(raw(default_value = r#""42""#))]
+        #[structopt(default_value = "42")]
```
TeXitoi: The test is now quite useless as the indent was to test raw default.

sphynx: Indeed. Removed.

PR #204, opened 2019-06-17
[Casing] Change default from verbatim to kebab.

0ndorio:
.. fixes TeXitoi/structopt#202

I hope I discovered all related comments and text snippets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
   as `Option<Vec<T>>` is handled differently. If you need to have a `Option<Vec<T>>`
   handled the old was, you can `type Something = Vec<T>;` and then use `Option<Something>`
   as your structopt field.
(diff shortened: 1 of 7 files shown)


TeXitoi:
Thanks. Will try to review today.

Please update the changelog.

0ndorio:
> Thanks. Will try to review today.
> 
> Please update the changelog.

Done. Listed it inside the `Breaking Changes` category.

TeXitoi:
Sorry to not been explicit enough: can you add a note in the changelog explaining how to have the same behavior as before, for ease of upgrading? Thanks.

0ndorio:
> Please improve the documentation at the begining of lib.rs, explaining that, by default, `long`, `short` `name` and commands without explicit name will be renamed using kebab-case if rename-all is not specified.
> 
> Else, everything looks good, thanks!

Updated the change log  and added an additional paragraph to describe the behavior (I hope it's in the right place).

TeXitoi:
Thanks!

PR #205, opened 2019-06-19
Use trybuild for testing expected proc macro errors

### LINKED ISSUE #199: Testing error messages from procedural macros
As noted [here](https://github.com/TeXitoi/structopt/pull/190#issuecomment-496913036), it would be nice to add tests for checking error messages got from procedural macros during compilation time. Since they are visible to users, it's important to keep them readable, predictable and stable between versions. 

An example of an error message from a macro in compile time:

```
  --> src/main.rs:11:10
   |
11 | #[derive(StructOpt, Debug)]
   |          ^^^^^^^^^
   |
   = help: message: cannot parse structopt attributes: ( parse ( 2 + 2 ) ): Error("expected identifier")

```

There is no easy way to run tests in compile time using conventional `[test]` attributes. However, I've found a nice package called [trybuild](https://github.com/dtolnay/trybuild) from @dtolnay, author of `syn` and `quote`. The package tries to compile test programs and matches the output against predefined `.stderr` files, which looks like exactly what is needed here.

I can take a closer look on that package and then incorporate it in `structopt` testing infrastructure.

sphynx:
Closes #199 

I tried to cover most of the "panics" from procedural macros in `attrs.rs` and to create at least one test for every panic.

By the way, those tests will be quite helpful to check whether the introduction of custom attributes parser (PR #198) preserves error messages well. Therefore, I propose to merge this PR first, then I'll rebase my parser changes from #198 on that and will update the tests or fix the error messages accordingly.


tests/ui/tuple_enum.stderr
```diff
@@ -0,0 +1,42 @@
+error[E0599]: no function or associated item named `augment_clap` found for type `u32` in the current scope
```
TeXitoi: lack the rs file, or this file should not be added

sphynx: Indeed, good catch. Fixed.

TeXitoi:
Thanks again!

PR #213, opened 2019-07-03
Support skipping struct fields

### LINKED ISSUE #174: Way to ignore some fields in struct
My config struct has a couple of fields that are unrelated to config parsing. It would nice to be able to mark them as ignored so no flags are generated for them and there's no requirement for the field to support FromStr and instead Require the field to implement Default.

sphynx:
Resolves #174

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@
 
 ## improvements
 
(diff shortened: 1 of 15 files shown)


TeXitoi:
Thanks!

PR #216, opened 2019-07-04
Add example for environment variables

### LINKED ISSUE #160: Add or_else to support arguments with environment fallbacks
I have a command-line program where some arguments are needed for the program to run, but can be read from environment variables if not explicitly set on the command line. At the moment, I'm using `clap` directly and writing:

```rust
let host = matches
        .value_of("host")
        .map(String::from)
        .or_else(|| env::var("HOST").ok())
        .expect("A hostname is required.");
```

What I'd prefer to write is:

```rust
#[derive(StructOpt)]
struct Arguments {
    #[structopt(short, long, or_else="host_from_env")]
    pub host: String;
}

fn host_from_env() -> Option<String> {
    env::var("HOST").ok()
}
```

The `or_else` would be a path of an in-scope function, and would be invoked if the value was not present in the clap arguments. If the `or_else` function returned `None`, the value would not be set and the arguments would be displayed as normal.

This approach would allow fallbacks besides just environment variables.

**Why not make the field optional and handle this outside `structopt`?**
Mostly for brevity/code clarity. The program requires that it have a hostname, it just allows for ways to provide it besides the command line. The second block of code above communicates that clearly, lets my fallback logic be clean and independent of the command-line parsing, and avoids needing to duplicate field names across two structs to support the "parsed-before-fallback" and "required-after-fallback" cases.

robinst:
Fixes #160.

### DIFF
--- a/examples/env.rs
+++ b/examples/env.rs
@@ -1,0 +1,23 @@
+use structopt::StructOpt;
+
+/// Example for allowing to specify options via environment variables.
+#[derive(StructOpt, Debug)]
+#[structopt(name = "env")]
+struct Opt {
+    // Use `env` to enable specifying the option with an environment
+    // variable. Command line arguments take precedence over env.
+    /// URL for the API server
+    #[structopt(long, env = "API_URL")]
+    api_url: String,
+
+    // The default value is used if neither argument nor environment
+    // variable is specified.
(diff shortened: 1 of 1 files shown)


TeXitoi:
Thanks!

PR #218, opened 2019-07-10
Explicitly specify the Result type

danielschemmel:
The name `Result` is evaluated in the user context, which may break when another `Result` type is used. For example, in the presence of [the basic error-chain quickstart](https://github.com/rust-lang-nursery/error-chain/blob/master/examples/quickstart.rs), this leads to:
```
   Compiling application v0.0.1 (/home/ghast/rust-cli-template)
error[E0107]: wrong number of type arguments: expected 1, found 2
  --> src/main.rs:13:10
   |
13 | #[derive(structopt::StructOpt, Debug)]
   |          ^^^^^^^^^^^^^^^^^^^^ unexpected type argument

For more information about this error, try `rustc --explain E0107`.
error: Could not compile `application`.
```

By explicitly specifying the `Result` type, this name clash can be resolved correctly.

### DIFF
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -437,7 +437,7 @@ fn gen_paw_impl(name: &Ident) -> TokenStream {
         impl paw::ParseArgs for #name {
             type Error = std::io::Error;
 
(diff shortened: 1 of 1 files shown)


TeXitoi:
Thanks

PR #221, opened 2019-07-15
Add documentation for `env`

vmalloc:


src/lib.rs
```diff
@@ -290,6 +290,23 @@
 //! # fn main() {}
 //! ```
 //!
+//! ## Environment Variable Fallback
+//!
+//! It is possible to specify an environment variable fallback option for arguments, so that their value is taken
+//! from the specified environment variables if not given through the command-line:
```
kaj: I think it should be "variable" in the singular here.

src/lib.rs
```diff
@@ -290,6 +290,23 @@
...
+//!
+//! ```
+//! # use structopt::StructOpt;
+//!
+//! #[derive(StructOpt)]
+//! #[structopt(name = "foo")]
+//! struct Foo {
+//!   #[structopt(short = "param", env = "PARAMETER_VALUE")]
```
kaj: The short form of an argument should generally be a single letter, so maybe `short = "p"`.  Or just use automatic naming like so:

```
 //!   #[structopt(long, short, env = "PARAMETER_VALUE")] 
```

kaj:
By the way, maybe the flag `hide_env_values` (useful to avoid printing out sensitive data from environment in help messages) should also be mentioned here.

src/lib.rs
vmalloc: 👍 fixed

src/lib.rs
vmalloc: Sorry - my bad. Fixed 👍

vmalloc:
@kaj I didn't know that one! Couldn't find it in the source though. Can you point me to a resource detailing it?

kaj:
> @kaj I didn't know that one! Couldn't find it in the source though. Can you point me to a resource detailing it?

It's listed at https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.hide_env_values (but the actual documentation is marked as TODO).  I use [it in a project](https://github.com/kaj/rphotos/blob/master/src/server/mod.rs#L57-L59) and it works fine.

vmalloc:
@kaj I'm not sure I get what that does... Using `env` by default does not output the value being used anywhere... what does that hide exactly?

kaj:
> @kaj I'm not sure I get what that does... Using `env` by default does not output the value being used anywhere... what does that hide exactly?

It does, in the `--help` output.  Try the example given in this crate:

```
env RETRIES=7 cargo run --example=env -- --help
```

vmalloc:
@kaj ooooh I see. Ok, will add it then

vmalloc:
@kaj I added documentation for `hide_env_values`

TeXitoi:
Thanks for the contribution!

Note that this is pure clap documentation, and I generally avoid to doubling this documentation. But `env` is a frequently asked feature, thus I'm OK to put it here.

Issue #217, opened 2019-07-04
Stop doing automatic author, about and version


TeXitoi:
Since the begining, structopt use the CARGO_PKG environment variables to populate version, about and authors if not provided. As sometime you don't want that to be setted, there is an ugly hack that, if you do `author = ""`, the magic is removed.

That's not so in par with the clap philosophy, as these setters are not called by default on raw clap.

Thus I propose to remove the ugly hack, don't use CARGO_PKG by default, but add without argument options to set these fiels using the CARGO_PKG, i.e.

```rust
#[derive(StructOpt)]
struct Opt {}
```
became
```rust
#[derive(StructOpt)]
#[structopt(about, author, version)]
struct Opt {}
```
and
```rust
#[derive(StructOpt)]
#[structopt(about = "", author = "", version = "")]
struct Opt {}
```
became
```rust
#[derive(StructOpt)]
struct Opt {}
```

To anyone reading this, what do you think?

Would fix #172 and maybe #173 

cc @sphynx @killercup @kbknapp @ssokolow @robinst @0ndorio

ssokolow:
I think the goal should be to default to a reasonable approximation of platform convention, similar to how Qt will try to default to matching the look of the target platform's GUI.

By that logic:

1. The current behaviour of `version` is perfectly reasonable as a default
2. The current behaviour of `about` is a footgun, because it competes with rustdoc for what purpose the struct's docstring should serve. (I just noticed *another* one of my projects slipped into having `--help` say `Command-line argument schema` because it decided to draw from the docstring that I include for the benefit of `rustdoc`.)
3. The current behaviour of `author` is at odds with GNU conventions, it's unusual enough that end-users are going to assume it's an opt-in decision and may mistake it for a sign that the structopt user is an egotist.

In short, I'd keep `version`'s current behaviour but, otherwise, I'm for this.

0ndorio:
I like the general idea to move into  saner defaults and I think the changes you proposed are a step into the right direction. So just ping me as soon as you decided what to do and I will help to implement it.

But I also think we will run into some odds based on the default behavior of clap and I think it's worth to push a bit more on their side to get rid of them instead of working on structopt internal fixes. This mostly includes the following topics:

 - **Meta Information**: Next to the version number and the author the default template output also contains the binary name on the first line. This might look okay as long as these information are combined but I think it will start to annoy as soon as we hide the version number and the author by default.
    
[...]

PR #225, opened 2019-08-12
Now error messages highlight the error location

CreepySkeleton:
A work on error reporting, as promised in #5 . 

First of all, I must genuinely apologize for this long delay, the summer had happened to be more busy and eventful than I'd expected. I don't think something like that will happen this month, although you never know what's going to happen, don't you? Sorry again.

About PR:

It seems like the current (old) implementation drops Spans as soon as possible which was quite sensible: there was no way to use it so there was no need to carry needless stuff around. Worse yet: most of data structures are `enum`s and one doesn't simply attach a piece of data to an `enum`. Basically, we have two ways: 

- first is to attach this data to every variant, like
  ```rust
  pub enum Ty {
      Bool(Span),
      Vec(Span),
      Option(Span),
      OptionOption(Span),
      OptionVec(Span),
      Other(Span),
  }
  ```
  Easy to see, this is extremely cumbersome. Despite this, I had to use this approach for some `enum`s 
  (or just one... can't remember), mostly because I was forced to by a `trait` interface.

- second is to separate the kind and the metadata:
  ```rust
  pub struct Ty {
      span: Span,
      kind: TyKind
  }
  
  pub enum TyKind { Bool, Vec /* ... */ }
  ```
  The drawback here is there are a quite a lot of `enum`s out here. I chose this approach, but slightly modified. There is `struct Sp<T>` in `spanned` module. It's essentially a wrapper around `T` which carries the span needed. For the sake of convince it implements `Deref` and some `From` traits.

Besides the `Span` problem mentioned, I've performed some refactoring, typically simplifying the code .

Most relevant changes are in `src/*.rs` files.

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -388,14 +388,15 @@ impl Attrs {
         {
             match segments.iter().last().unwrap().ident.to_string().as_str() {
                 "bool" => t(Ty::Bool),
-                "Option" => match sub_type(ty).map(Attrs::ty_from_field) {
-                    Some(ty) => match *ty {
+                "Option" => sub_type(ty)
+                    .map(Attrs::ty_from_field)
+                    .map(|ty| match *ty {
                         Ty::Option => t(Ty::OptionOption),
                         Ty::Vec => t(Ty::OptionVec),
                         _ => t(Ty::Option),
-                    },
-                    None => t(Ty::Option),
-                },
+                    })
+                    .unwrap_or(t(Ty::Option)),
+
                 "Vec" => t(Ty::Vec),
                 _ => t(Ty::Other),
             }
@@ -419,7 +420,7 @@ impl Attrs {
                 }
                 if !res.methods.is_empty() {
                     span_error!(
-                        res.methods[0].name.span(),
+                        res.kind.span(),
                         "methods and doc comments are not allowed for flattened entry"
                     );
                 }
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -5,23 +5,26 @@
 // <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
+use crate::spanned::Sp;
 use heck::{CamelCase, KebabCase, MixedCase, ShoutySnakeCase, SnakeCase};
 use proc_macro2::{Span, TokenStream};
+use proc_macro_error::span_error;
 use quote::quote;
 use std::{env, mem};
+use syn::spanned::Spanned as _;
 
 use syn::Type::Path;
 use syn::{
-    self, AngleBracketedGenericArguments, Attribute, GenericArgument, Ident, MetaNameValue,
+    self, AngleBracketedGenericArguments, Attribute, GenericArgument, Ident, LitStr, MetaNameValue,
     PathArguments, PathSegment, TypePath,
 };
 
 use crate::parse::*;
 
-#[derive(Copy, Clone, PartialEq, Debug)]
+#[derive(Clone, Debug)]
 pub enum Kind {
-    Arg(Ty),
-    Subcommand(Ty),
+    Arg(Sp<Ty>),
+    Subcommand(Sp<Ty>),
     FlattenStruct,
     Skip,
 }
@@ -36,20 +39,8 @@ pub enum Ty {
     Other,
 }
 
-#[derive(Debug)]
-pub struct Attrs {
-    name: String,
-    cased_name: String,
-    casing: CasingStyle,
-    methods: Vec<Method>,
-    parser: (Parser, TokenStream),
-    has_custom_parser: bool,
-    kind: Kind,
-}
-
-#[derive(Debug)]
-struct Method {
-    name: String,
+pub struct Method {
+    name: Ident,
     args: TokenStream,
 }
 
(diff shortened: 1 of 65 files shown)


TeXitoi:
Thanks for the contribution. I'll try to be fast on the review, but no promise.

Don't be sorry! We give our free time on these project, and thus we don't have any obligation.

structopt-derive/src/attrs.rs
```diff
@@ -341,155 +356,207 @@ impl Attrs {
...
+                "bool" => t(Ty::Bool),
                 "Option" => match sub_type(ty).map(Attrs::ty_from_field) {
-                    Some(Ty::Option) => Ty::OptionOption,
-                    Some(Ty::Vec) => Ty::OptionVec,
-                    _ => Ty::Option,
+                    Some(ty) => match *ty {
+                        Ty::Option => t(Ty::OptionOption),
+                        Ty::Vec => t(Ty::OptionVec),
+                        _ => t(Ty::Option),
+                    },
```
TeXitoi: minor: I think you can map to flatten this match

structopt-derive/src/parse.rs
```diff
@@ -130,22 +120,13 @@ impl Parse for ParserSpec {
...
-    }
-    s_opt_attrs
+    all_attrs
+        .iter()
+        .filter(|attr| attr.path.is_ident("structopt"))
+        .map(|attr| {
+            let attrs: StructOptAttributes = parse2(attr.tts.clone()).unwrap_or_exit();
+            attrs.attrs
+        })
+        .flatten()
```
TeXitoi: map+flatten = flat_map?

tests/ui/flatten_and_doc.stderr
```diff
@@ -1,7 +1,5 @@
-error: proc-macro derive panicked
+error: methods and doc comments are not allowed for flattened entry
```
TeXitoi: Can't we point to the flatten or doc comment??

tests/ui/skip_without_default.rs
```diff
@@ -23,7 +23,4 @@ pub struct Opts {
     k: Kind,
 }
 
-fn main() {
-    let opts = Opts::from_args();
-    println!("{:?}", opts);
-}
+fn main() {}
```
TeXitoi: No span modification here?

tests/ui/structopt_empty_attr.stderr
```diff
@@ -1,7 +1,5 @@
-error: proc-macro derive panicked
+error: unexpected end of input, expected parentheses
```
TeXitoi: Not possible to have span on the attribute here?

tests/ui/skip_without_default.rs
CreepySkeleton: I'm not sure what you mean by "span modification" here, if it's about the reason why I emptied the `main`'s body - well, before this PR the macro was reporting error by a plain panic. Such panics abort compilation process effectively immediately, so the panic-error is the only error displayed *even if there are other errors*. But now errors are reported by a token stream containing a `compile_error!` invocation and such invocation *does not* abort a compilation right away. Furthermore, any `impl Structopt for Foo` code was not generated, so we're going to have a `Foo does not have "from_args" method` error in addition to our `compile_error!` invocation, and this is not what we test here. So I decided to keep this stuff as clean and simple as possible.

structopt-derive/src/parse.rs
CreepySkeleton: Right

tests/ui/skip_without_default.rs
TeXitoi: Sorry, I was really not clear.

This particular test has the same error message as before, that is
```
error[E0277]: the trait bound `Kind: std::default::Default` is not satisfied
  --> $DIR/skip_without_default.rs:17:10
   |
17 | #[derive(StructOpt, Debug)]
   |          ^^^^^^^^^ the trait `std::default::Default` is not implemented for `Kind`
   |
   = note: required by `std::default::Default::default`

For more information about this error, try `rustc --explain E0277`.
```

Somewhere, in the generated code, we must do something like that:
```rust
k: Default::default(),
```
without any span information. Thus, the error message is at the call site. It would be great to point to the `k` field of the struct. If it imply complicated modification, never mind.

tests/ui/structopt_empty_attr.stderr
CreepySkeleton: Done. Not the cleanest code I've ever wrote, by our options are limited here by `trait syn::Parse` interaface

tests/ui/skip_without_default.rs
CreepySkeleton: Will do

structopt-derive/src/attrs.rs
CreepySkeleton: Done

tests/ui/skip_without_default.rs
CreepySkeleton: Wow, that was really easy, ~2mins total. Apparently my idea of carrying as much as possible span info around proved very useful :)

tests/ui/flatten_and_doc.stderr
CreepySkeleton: Done, now we point to the flatten

TeXitoi:
Thanks! Now I must find time to do #217 and then publish v0.3.

CreepySkeleton:
You should consider to publish v0.2.19 so people who have to stick to v0.2.x for one reason or another could make use of this improvement. Contrary wise, you may want not to do that to encourage people to move to v0.3.x  and publish 0.2.19 only if/when requested. Anyway, you should know better.

TeXitoi:
There is a lot to backport for that. Current master is already breaking 0.2.

PR #227, opened 2019-08-15
Update dependencies

CreepySkeleton:
`syn`, `proc_macro2` and `quote` were updated recently. Since you're going to publish a breaking version anyway we could ship this update along. Anyway, we will have to do it sooner or later and now it's a perfect opportunity.


TeXitoi:
Any idea why the span of `Option<Option<T>>` is different on nightly? (that's why travis is not happy)

TeXitoi:
Your git config seems also unsync with your git account.

CreepySkeleton:
> Your git config seems also unsync with your git account.

Oops, thanks for pointing out :)

> Any idea why the span of `Option<Option<T>>` is different on nightly? (that's why travis is not happy)

It's about improvements in `proc-macro2::Span`: earlier it haven't had all the info necessary to highlight all the parts of a complex type at once on stable but it seems they now have some magic on nightly that allows it. 

We have multiple ways to address this issue 
- We could test `cargo build` but not `cargo test` on nightly - if it works on stable it *probably* should work on nightly. We may roll things back once they're settled out.
- We could use 
  ```
  allow_failures:
      - rust: nightly
```
  in `travis.yml` and roll back when it's fixed.
- We could make different test cases for stable and nightly - waste of efforts if you ask me, but I still would do it if you want.

I leave it at your discretion

**Comment by CreepySkeleton** [comment]
@TeXitoi So what do de do about it? I suggest `allow_failure` or "separate test suits for stable and nightly" solutions: both are easy to implement and unroll when needed

**Comment by TeXitoi** [comment]
Sorry. Can we do failure only on stable? That would be perfect. Else, let's do tests only on stable, as it will pose problems later when the release train change if we just allow failure on nightly.

**Comment by TeXitoi** [comment]
"failure only on stable" I mean ui tests only on stable.

**Comment by CreepySkeleton** [comment]
I'm not sure I understand you correctly, are you're saying you want not to run ui tests on nightly but run all the other tests? If yes, that's exactly what's in this PR now.

**Comment by CreepySkeleton** [comment]
@TeXitoi Unfortunately, this is not all sunshine and bubbles. `clap` implicitly depends on [`failure`](https://github.com/rust-lang-nursery/failure) and `failure_derive` crates, and they in turn depend on old (pre 1.*) versions of `quote/proc_macro2/syn` crates (they **are** in [the process of updating](https://github.com/rust-lang-nursery/failure/pull/319) though). So now we have to double-build `syn/proc_macro2/quote` crates. Let's wait for a few days before merging this (I'll poke these guys tomorrow and I'll offer my help to speed up things)

**Comment by CreepySkeleton** [comment]
By the way, it seems like `nightly` feature is obsolete:

https://github.com/alexcrichton/proc-macro2/blob/75972f61403c6cc453608f3358cbfdef6edea4b5/Cargo.toml#L39

**Comment by TeXitoi** [comment]
We can remove the nightly feature.

File: tests/macro-errors.rs [hunk 1]
```diff
@@ -5,8 +5,15 @@
...
 
+extern crate version_check;
+use version_check::Channel;
+
 #[test]
 fn ui() {
-    let t = trybuild::TestCases::new();
-    t.compile_fail("tests/ui/*.rs");
+    if Channel::read().unwrap().is_nightly() {
+        eprintln!("nightly build detected, skipping ui tests");
```
TeXitoi: When this nightly will go to beta, beta will also fail :-/

CreepySkeleton: And when beta go stable we're going to have a crash on stable \***bad sob**\*. Well, are we going to skip beta too?

TeXitoi: I don't know what to do, but I think at the end, the test will be done everywhere except on the smallest supported rust version.

TeXitoi: Maybe we can for the moment have nightly and beta as allow_failure, and when stable will break, we will update the test, remove the allow failure on stable and nightly, and remove the tests on the smallest supported rust version (i.e. this version will just compile)

TeXitoi: I don't like that, but I don't have a better proposition.

CreepySkeleton: Honestly, the more I think about the idea of skipping tests on certain channels the less I like it. I'm not sure what your policy on supporting nightly is, but I guess from the existence of `nightly` feature you claim to provide at least partial support, i.e. "it works, until these compiler guys break something again". Such a support implies that the errors that *must* happen will  actually happen.

If we chose to ignore these tests we effectively say "sorry, nightly guys, but we don't check that all the parts are *actually* work", and while this still would work it's not the best support we can provide. Worse yet, should something *actually* break in ui-tests - we won't even know about it.

I propose one of these choices instead
- Separate test suits for every channel, i.e `ui-stable/ui-beta/ui-nightly` subfolders. Only 4/30 errors are different, it's not too much to look after. The main drawback here - we'll have to support additional tests, at least until `stable/nightly/beta` are in sync. The main advantage - should something actually break and we know about it the very instant.
- We may use `allow_failures`, so travis won't bother you with this ui stuff and we still will be able to periodically check whether the error has changed. Drawback - we won't be notified, we'll have to *check* manually. (The question - is there a way to configure travis to notify once error changed? That would have annihilate the drawback).

What do you think?

TeXitoi: Can't we switch the .stderr file according to the rust version?

CreepySkeleton: Exactly my first proposal - pick tests depending on channel, details may differ: 
- separate folders
- separate folders + symlinks
- explicit code in `macro_errors.rs` that picks needed stuff

I'm about to make few commits so you could judge for yourself

CreepySkeleton: And [here is](https://github.com/CreepySkeleton/structopt/tree/update_deps2) another option

TeXitoi: * is it possible without duplication the input files?
* can't we switch on the rust version (>= 1.39, < 1.39) instead of the channel?
* If duplication is not possible, can we have a test suite of common, and another spited by version

CreepySkeleton: >     * is it possible without duplication the input files?

Symlinks?

CreepySkeleton: >     * can't we switch on the rust version (>= 1.39, < 1.39) instead of the channel?

Good idea, we definitely can

CreepySkeleton: >     * If duplication is not possible, can we have a test suite of common, and another spited by version

[Like this?](https://github.com/CreepySkeleton/structopt/tree/update_deps2)

TeXitoi: The `update_deps2` version seems perfect, with 2 sets switched by rust version.

CreepySkeleton:
@TeXitoi This seems complete. You can merge it once people manage to update their deps

TeXitoi:
As removing the nightly feature is a breaking change, we must merge this before 0.3. I don't really care if for now the user need to compile 2 times syn and quote as using the 1 version is the thing to do and the update of the deps will be transparent.

Can you update the changelog about the removal of the nightly feature, and I'll merge this.

TeXitoi:
I'm OK to merge this now if you update the changelog in the other PR.

TeXitoi:
Thanks!

PR #229, opened 2019-08-21
Change `author/about/version` behavior

### LINKED ISSUE #217: Stop doing automatic author, about and version

CreepySkeleton:
Closes #217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,16 +3,27 @@
 ## Breaking changes
 
 * Bump minimum rust version to 1.34 by [@TeXitoi](https://github.com/TeXitoi)
+
 * Support optional vectors of arguments for distinguishing between `-o 1 2`, `-o` and no option provided at all
   by [@sphynx](https://github.com/sphynx)
   ([#180](https://github.com/TeXitoi/structopt/issues/188)). This is a breaking change
   as `Option<Vec<T>>` is handled differently. If you need to have a `Option<Vec<T>>`
   handled the old was, you can `type Something = Vec<T>;` and then use `Option<Something>`
   as your structopt field.
+
 * Change default case from 'Verbatim' into 'Kebab' by [@0ndorio](https://github.com/0ndorio)
   ([#202](https://github.com/TeXitoi/structopt/issues/202)). This is a breaking change.
   If you rely on the old behavior you need to add `#[structopt(rename_all = "verbatim")]`
   as an attribute to each data structure deriving the `StructOpt` trait.
+
+* Change `version`, `author` and `about` attributes behavior. `version/author/about = ""` is no longer a
+  valid syntax. `author` and `about` are no longer derived from `Cargo.toml` by default, i.e when no
+  attributes mentioned. `version` is still to be derived from `Cargo.toml` by default.
+  Introduced explicit `author` and `about` attributes (with no arguments, i.e `#[structopt(author)]`)
+  for explicit requests to deduce author/about fields from `Cargo.toml`.
+  `version/author/about = "version/author/about"` is still valid syntax.
+  Introduced `no_version` attribute (no args) which prevents version auto deducing by default.
+  Proposed by [@TeXitoi](https://github.com/TeXitoi) [(#217)](https://github.com/TeXitoi/structopt/issues/217), implemented by [@CreepySkeleton](https://github.com/CreepySkeleton) [(#229)](https://github.com/TeXitoi/structopt/pull/229).
 
 ## improvements
 
@@ -21,6 +32,9 @@
 
 * Add optional feature to support `paw` by [@gameldar](https://github.com/gameldar)
   ([#187](https://github.com/TeXitoi/structopt/issues/187))
+
+* Significantly improve error reporting by [@CreepySkeleton](https://github.com/CreepySkeleton)
+  ([#225](https://github.com/TeXitoi/structopt/pull/225/))
 
 # v0.2.16 (2019-05-29)
 
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -244,7 +244,7 @@ impl Attrs {
 
                 MethodCall(name, args) => self.methods.push(Method {
                     name: name.into(),
-                    args: quote!(#args),
+                    args: quote!(#(#args),*),
                 }),
 
                 RenameAll(_, casing_lit) => {
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -5,18 +5,18 @@
 // <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
+
 use crate::spanned::Sp;
+
+use std::env;
+
 use heck::{CamelCase, KebabCase, MixedCase, ShoutySnakeCase, SnakeCase};
 use proc_macro2::{Span, TokenStream};
-use proc_macro_error::span_error;
+use proc_macro_error::{call_site_error, span_error};
 use quote::quote;
-use std::{env, mem};
-use syn::spanned::Spanned as _;
-
-use syn::Type::Path;
 use syn::{
-    self, AngleBracketedGenericArguments, Attribute, GenericArgument, Ident, LitStr, MetaNameValue,
-    PathArguments, PathSegment, TypePath,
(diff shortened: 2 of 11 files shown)


CreepySkeleton:
I decided not to create a new PR for the `raw` check, I affects the same lines that the first commit does, I'd have to stack my work on top of it anyway

CreepySkeleton:
Interesting...

CreepySkeleton:
@TeXitoi Could you please take a look? It seems like travis crashes on *all* channels with some compiler-bug error. I can't reproduce it locally, my `cargo test` is flawless and shining

CreepySkeleton:
[My travis build](https://travis-ci.org/CreepySkeleton/structopt/builds/575687606) can't reproduce it. Any idea?

TeXitoi:
Strange...

CreepySkeleton:
Yahoo! Now travis is all green just like a troll from 4chan. Can I ask you to download and share build artifacts of [this particular build](https://travis-ci.org/TeXitoi/structopt/builds/575667984?utm_source=github_status&utm_medium=notification) with me? I'd like to dig into this bug or at least report it to rustc team

TeXitoi:
I can't find anything to download. Any pointer?

CHANGELOG.md
```diff
@@ -3,17 +3,30 @@
 ## Breaking changes
 
 * Bump minimum rust version to 1.34 by [@TeXitoi](https://github.com/TeXitoi)
+
+* Update `syn`, `proc_macro2` and `quote` dependencies to version `1`.
```
TeXitoi: That's for the other PR, no?

structopt-derive/src/attrs.rs
```diff
@@ -536,16 +560,60 @@ impl Attrs {
...
+    }
+
+    /// generate methods on top of a field
+    pub fn field_methods(self) -> TokenStream {
+        let methods: Vec<_> = self
             .methods
             .iter()
-            .map(|&Method { ref name, ref args }| quote!( .#name(#args) ));
+            .map(|&Method { ref name, ref args }| quote!( .#name(#args) ))
+            .collect();
```
TeXitoi: I think you don't need to collect, quote take an IntoIterator.

structopt-derive/src/attrs.rs
```diff
@@ -536,16 +560,60 @@ impl Attrs {
...
+
+        self.about
+            .clone()
+            .map(|(ident, arg)| self.push_str_method(ident.into(), arg.into()));
+
+        let methods: Vec<_> = self
+            .methods
+            .iter()
+            .map(|&Method { ref name, ref args }| quote!( .#name(#args) ))
+            .collect();
```
TeXitoi: Same here

structopt-derive/src/lib.rs
```diff
@@ -282,7 +283,7 @@ fn gen_clap(attrs: &[Attribute]) -> GenOutput {
     let attrs = Attrs::from_struct(attrs, Sp::call_site(name), Sp::call_site(DEFAULT_CASING));
     let tokens = {
         let name = attrs.cased_name();
-        let methods = attrs.methods();
+        let methods = attrs.clone().top_level_methods();
```
TeXitoi: Why are these clone needed now?

CreepySkeleton:
I don't know, I can't find anything that 

> I can't find anything to download. Any pointer?

I don't know, I can't find any way to do that in google so [I asked](https://stackoverflow.com/questions/57621915/is-there-a-way-to-download-compilation-artifacts-from-a-travis-ci-build) the overmind, let's see what we can do

CHANGELOG.md
CreepySkeleton: True, I'll get rid of it

structopt-derive/src/attrs.rs
CreepySkeleton: That is correct, the `.collect` is going to be removed

structopt-derive/src/attrs.rs
CreepySkeleton: Yes, I'll make it perish

structopt-derive/src/lib.rs
CreepySkeleton: i just realized that this is quite complicated question, let me think...

hellow554:
> 
> 
> [My travis build](https://travis-ci.org/CreepySkeleton/structopt/builds/575687606) can't reproduce it. Any idea?

Looks like an incremental bug. I think it is covered in https://github.com/rust-lang/rust/issues/63161, you can leave a comment there.

CreepySkeleton:
> 
> 
> > [My travis build](https://travis-ci.org/CreepySkeleton/structopt/builds/575687606) can't reproduce it. Any idea?
> 
> Looks like an incremental bug. I think it is covered in [rust-lang/rust#63161](https://github.com/rust-lang/rust/issues/63161), you can leave a comment there.

Hmm, I like to leave comments, *especially* on my second vacation day :-)

structopt-derive/src/lib.rs
CreepySkeleton: In my previous attempt these methods mutated `self`. Hence we happen to reuse `attrs` here first call would add methods that are acceptable for `clap::App` but not for `clap::Arg`. First we used this to generate `App` and later we generated `Arg`. My current approach is different, these methods mutate self no longer.

CreepySkeleton:
Also, I squashed all the commits into one, tell me if you want to separate them

TeXitoi:
I squash at merge time. That's simpler if I can review the modifications added, thanks.

CreepySkeleton:
May I ask why do you always squash?

TeXitoi:
One commit, one feature. And some contributors do a lot of commit as "fix typo", "fmt", "oups, bug fix" that can be annoying in a git bisect.

CreepySkeleton:
Yeah, I do understand why people squash meaningless commits like `oops`, `penis` and stuff - but I also find useful to separate related but different things of one giant commit into separate ones. For example, if it was me in your shoes I'd probably break this commit into three - `Change behavior...`, `Add \`raw\` removal warning` and `Update Changelog`. Sometimes feature implementing may grow into one pretty monstrous commit and it's easier to shatter not-really-related things into smaller peaces. IMHO of course...

TeXitoi:
If you do that, I'll just rebase, not squash

TeXitoi:
I don't want to put pressure on contributors about managing commits in a certain way, but I might ask for a separate PR for unrelated stuff. I allowed you to do one PR for raw and change behavior, thus I will not ask you to do 2 PR, but if you separate this work in 2 commit, I'll happily
keep the history with these 2 commits. I will not take the time to separate these 2 commits myself.

CreepySkeleton:
> but if you separate this work in 2 commit, I'll happily
> keep the history with these 2 commits. I will not take the time to separate these 2 commits myself.

Exactly what I would do, except I'd probably just isolate these 2 commits myself. 

N.B: I'm not actually arguing, I just noticed some policy that is different from mine and got curious about it's advantages and drawbacks. That's how people learn, and I'm trying.

TeXitoi:
Thanks!

CreepySkeleton:
Mm, @TeXitoi did you look at travis output? I'm asking because it seems we missed a warning, I just was about to push the fix when you merged it :)

TeXitoi:
Sorry no, just looked at the status

sourcefrog:
It seems like the docs are out of date with this? They seem to still indicate, in https://docs.rs/structopt/0.3.14/structopt/#magical-methods that these default to the Cargo values.

Issue #233, opened 2019-08-25
How to implement repeated arguments ?


lovasoa:
clap distinguishes between multiple values and multiple occurrences:
https://docs.rs/clap/2.20.3/clap/struct.Arg.html#method.multiple

Is it possible to have an argument that takes multiple occurrences and a single value for each occurrence in structopt ?

lovasoa:
Here is what I tried so far, but does not work : 

```rust
use structopt::StructOpt;

#[derive(StructOpt, Debug)]
pub struct Arguments {
    input_uri: Option<String>,
    #[structopt(short = "H", long = "header", parse(try_from_str = "parse_header"), number_of_values = 1)]
    headers: Vec<(String, String)>,
}

fn parse_header(s: &str) -> Result<(String, String), &'static str> {
    let vals: Vec<&str> = s.splitn(2, ':').map(str::trim).collect();
    if let [key, value] = vals[..] {
        Ok((key.into(), value.into()))
    } else {
        Err("Invalid header format. Expected 'Name: Value'")
    }
}

#[test]
fn test_headers_and_input() -> Result<(), structopt::clap::Error> {
    let args: Arguments = StructOpt::from_iter_safe([
        "--header", "Referer: http://test.com",
        "--header", "User-Agent: custom",
        "--header", "A:B",
        "input-url"
    ].iter())?;
    assert_eq!(args.input_uri, Some("input-url".into()));
    assert_eq!(args.headers, vec![
        ("Referer".into(), "http://test.com".into()),
        ("User-Agent".into(), "custom".into()),
        ("A".into(), "B".into()),
    ]);
    Ok(())
}
```

TeXitoi:
```rust
#[structopt(short, nimber_of_values =1)]
i: Vec<String>
```

On mobile so not tested.

lovasoa:
Oh, my bad, `from_iter_safe` takes the name of the program as the first element of the iterator. My test was faulty : 

```rust
#[test]
fn test_headers_and_input() -> Result<(), structopt::clap::Error> {
    let args: Arguments = StructOpt::from_iter_safe([
        "dezoomify-rs",
        "--header", "Referer: http://test.com",
        "--header", "User-Agent: custom",
        "--header", "A:B",
        "input-url"
    ].iter())?;
    assert_eq!(args.input_uri, Some("input-url".into()));
    assert_eq!(args.headers, vec![
        ("Referer".into(), "http://test.com".into()),
[...]

PR #234, opened 2019-08-25
Update keyvalue example

lovasoa:
See #233


lovasoa:
@TeXitoi : is that ok ?

TeXitoi:
It was night. That's OK, merged.

PR #246, opened 2019-08-31
Fix #245

### LINKED ISSUE #245: methods are not allowed for skipped fields when using multi-line doc comments
Title. Structopt 0.3.0 errors out with `methods are not allowed for skipped fields` if skipped fields have doc comments, but only multi-line ones.

Repro:

```rust
#[derive(StructOpt)]
struct Example {
    /// This does an internal thing.
    /// Run-on sentence explanation.
    #[structopt(skip)]
    works: String,

    /// This does another internal thing.
    ///
    /// # IMPORTANT NOTE
    #[structopt(skip)]
    fails: String,
}
```

CreepySkeleton:
Closes #245

### DIFF
--- a/tests/skip.rs
+++ b/tests/skip.rs
@@ -92,3 +92,35 @@ fn skip_enum() {
         }
     );
 }
+
+#[test]
+fn skip_help_doc_comments() {
+    #[derive(StructOpt, Debug, PartialEq)]
+    #[structopt(name = "a")]
+    pub struct Opt {
+        #[structopt(skip, help = "internal_stuff")]
+        a: u32,
+
+        #[structopt(skip, long_help = "internal_stuff\ndo not touch")]
+        b: u32,
+
+        /// Not meant to be used by clap.
+        ///
+        /// I want a default here.
+        #[structopt(skip)]
+        c: u32,
+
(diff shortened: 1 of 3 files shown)


CreepySkeleton:
This is just a fast patch. I'm working on a big refactoring that introduces explicit AST and separates input processing into phases: `parse`, `validate`, `generate`. Current implementation is almost unworkable when it comes to parent - children relationships

tests/skip.rs
```diff
@@ -92,3 +92,34 @@ fn skip_enum() {
...
+    #[structopt(name = "a")]
+    pub struct Opt {
+        #[structopt(skip, help = "internal_stuff")]
+        a: u32,
+
+        #[structopt(skip, long_help = "internal_stuff\ndo not touch")]
+        b: u32,
+
+        /// Not meant to be used by clap.
+        /// I want a default here.
```
TeXitoi: Did you check the test fail without the fix? Because I think an empty line is needed at the second line to generate a long_help

CreepySkeleton: Yep, forgot to change

TeXitoi:
Thanks

PR #248, opened 2019-09-02
Emit dummy impls on error

### LINKED ISSUE #240: Emit dummy StructOpt impl when reporting errors
Currently when structopt-derive reports an error, only the compile_error is emitted. That means the author will see both the compile_error *and* most likely an error about a missing StructOpt impl.

```console
error: `CARGO_PKG_VERSION` environment variable is not defined, use `version = "version" to set it manually or `no_version` to not set it at all
 --> src/main.rs:5:10
  |
5 | #[derive(StructOpt, Debug)]
  |          ^^^^^^^^^

error[E0599]: no function or associated item named `from_args` found for type `Opt` in the current scope
  --> src/main.rs:44:20
   |
7  | struct Opt {
   | ---------- function or associated item `from_args` not found for this
...
44 |     let opt = Opt::from_args();
   |                    ^^^^^^^^^ function or associated item not found in `Opt`
   |
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `from_args`, perhaps you need to implement it:
           candidate #1: `structopt::StructOpt`

error: aborting due to 2 previous errors
```

It would be nicer to emit a dummy StructOpt impl in addition to the error. Nothing in the impl will be called because the crate doesn't compile from the compile_error, but this way the author gets less useless errors.

CreepySkeleton:
Closes #240 

Also I moved `tests/ui` to `tests/ui-common`.

I'm not sure what to write in changelog, could use your guidance here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # v0.3.1 (2019-08-31)
 * Fix error messages ([#241](https://github.com/TeXitoi/structopt/issues/241))
 * Fix "`skip` plus long doc comment" bug ([#245](https://github.com/TeXitoi/structopt/issues/245))
+* Now `structopt` emits dummy `StructOpt` implementation along with an error. It suppresses
+  meaningless errors like `from_args method is not found for Opt`
 
 # v0.3.0 (2019-08-30)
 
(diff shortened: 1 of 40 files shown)


CreepySkeleton:
Changelog updated + rebased. Ready to go in

TeXitoi:
Thanks!

PR #250, opened 2019-09-03
Fix changelog example

### LINKED ISSUE #249: Changelog for optional vectors may have wrong example for old behaviour
From the changelog; issue in comments:

> If you need to fall back to the old behavior you can use a type alias:
> 
> ```rust
> type Something = Vec<String>; // Not actually used
> #[derive(StructOpt)]
> struct Opt {
>   #[structopt(long)]
>   fruit: Option<Vec<String>>, // Shouldn't this be `Option<Something>`?
> }
> ```

CreepySkeleton:
Closes #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ type Something = Vec<String>;
 #[derive(StructOpt)]
 struct Opt {
   #[structopt(long)]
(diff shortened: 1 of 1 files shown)


TeXitoi:
Thanks

PR #251, opened 2019-09-05
Do not call .verison() when CARGO_PKG_VERSION is not set

### LINKED ISSUE #243: Accept rustflag equivalent to no_version
Structopt currently looks at the environment variable $CARGO_PKG_VERSION to determine a version for the command line app. In non-Cargo environments this environment variable doesn't make sense and we would prefer to use no_version by default everywhere.

I would recommend for structopt-derive to accept a rustflag cfg that disables the check for CARGO_PKG_VERSION and assumes no_version by default.

```shell
RUSTFLAGS='--cfg structopt_no_version' cargo build ...
```

If a `version = "..."` attribute is present, that should take precedence over structopt_no_version.

Implementation-wise this would look like `#[cfg(structopt_no_version)]` and `#[cfg(not(structopt_no_version))]` codepaths in structopt-derive.

cc @bolinfest

CreepySkeleton:
It looks like we can arguing on #243 just forever. 

Docs are upcoming

fixes #243

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Fix "`skip` plus long doc comment" bug ([#245](https://github.com/TeXitoi/structopt/issues/245))
 * Now `structopt` emits dummy `StructOpt` implementation along with an error. It suppresses
   meaningless errors like `from_args method is not found for Opt`
+* `.version()` not get generated if `CARGO_PKG_VERSION` is not set anymore.
 
 # v0.3.0 (2019-08-30)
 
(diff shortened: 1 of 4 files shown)


TeXitoi:
Waiting for the doc.

CreepySkeleton:
Written and redacted by me the magnificent! I believe docs are in sync with current implementation, more or less. Could use some future proofreading though.

dtolnay:
Thanks! Would it be possible to get an 0.3.1 release with this change?

TeXitoi:
wait a few seconds, I'm doing it ;-)

TeXitoi:
v0.3.1 published

PR #263, opened 2019-09-19
Suppress all clippy warnings

CreepySkeleton:
Closes #262 

Now we put `#[allow(clippy)]` on top of every `impl` block, suppressing **all** clippy warnings, not just select few. In my opinion, any attempt to generate "flawless" code is meaningful and useless.

PR #264, opened 2019-09-19
Fix clippy warnings

BeatButton:
Even after #263 I'm still getting some warnings.
```
warning: called `map(f).unwrap_or_else(g)` on an Option value. This can be done more directly by calling `map_or_else(g, f)` instead
   --> src\main.rs:167:15
    |
167 |         urls: Vec<String>,
    |               ^^^
    |
    = note: replace `map(Vec).unwrap_or_else(Vec)` with `map_or_else(Vec, Vec)`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unwrap_or_else
```
I think this one is probably just worth fixing?
Tests all pass still, of course, since this doesn't change any behavior at all.

### DIFF
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -290,8 +290,7 @@ fn gen_constructor(fields: &Punctuated<Field, Comma>, parent_attribute: &Attrs) 
                     Ty::OptionVec => quote_spanned! { ty.span()=>
                         if matches.is_present(#name) {
                             Some(matches.#values_of(#name)
-                                 .map(|v| v.map(#parse).collect())
-                                 .unwrap_or_else(Vec::new))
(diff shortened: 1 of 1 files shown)


CreepySkeleton:
Do you test it on v0.3.2? #263 will be released in v0.3.3.

Still, I'm agree with this change, let's have it

CreepySkeleton:
@BeatButton I could really use your reply because I thought `#[allow(clippy)]` in #263 should suppress all the possible clippy warnings. It isn't a fix if it doesn't. Could you past a reproducer so we could check it?

BeatButton:
I tested it against the master branch at the time I made the pull request.
The code I changed was in the function ``get_constructor`` in ```structopt-derive```, which was not annotated with ```#[allow(clippy)]```, and probably shouldn't be.
I just confirmed by checking against #263.

PR #271, opened 2019-10-09
Add from_flag parser

### LINKED ISSUE #185: Support AtomicBool
I wanted to replace a `bool` in a struct opt with an `AtomicBool` and this didn't "just work". I suppose I have to write a custom parser for this, and pass it as a `parse(...)`, but it would be nice if that would just work.

rokob:
Fixes #185 

This PR adds the `from_flag` parser type which allows non bool field types to be treated as flags.


TeXitoi:
You also need a rustfmt.

TeXitoi:
Thanks

TeXitoi:
v0.3.3 published.

PR #272, opened 2019-10-09
Run ui tests only on stable

CreepySkeleton:
Well, it looks like nightly `Span` improvements are not going to be stabilized any time soon.
I propose to get back to `stable/nightly` split instead of `pre.1.39/post.1.39`.

renamed folders:
* `ui-pre_1.39` => `ui-stable-beta`
* `ui-1.39_post` => `ui-nightly`


CreepySkeleton:
Nope, beta tests would require a separate folder.

I just proposed [a new trybuild feature](https://github.com/dtolnay/trybuild/issues/36) that should allow to easily resolve this embarrassment. We'll see what will come out of this.

TeXitoi:
Thanks!

PR #275, opened 2019-10-18
docs: observes that the trait is more than from_args

### LINKED ISSUE #107: Access to `usage` for custom error messages
In some cases, I need to do validation after-the-fact..  With clap, I have access to `ArgMatches` which can let be print the usage.  I'm not aware of a way to access that from structopt.

thiagoarrais:
Fixes #107 

Simply adds an observation to the docs stating that the `StructOpt` trait is more than `from_args`.

Please feel free to point any mistakes. Especially in my English, since it isn't my first language.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -104,6 +104,11 @@
 //! (unless they're [subcommands](#subcommands)),
 //! and you can adjust these apps and args by `#[structopt(...)]` [attributes](#attributes).
 //!
+//! The derived `StructOpt` trait allows access to other methods besides
+//! `from_args`. You can, for instance, get the underlying `clap::ArgMatches`
+//! by doing `Opt::clap().get_matches()`. See the [trait's reference documentation](trait.StructOpt.html)
+//! for more details.
+//!
 //! ## Attributes
 //!
 //! `#[structopt(...)]` attributes fall into two categories:


src/lib.rs
```diff
@@ -104,6 +104,11 @@
 //! (unless they're [subcommands](#subcommands)),
 //! and you can adjust these apps and args by `#[structopt(...)]` [attributes](#attributes).
 //!
+//! The derived `StructOpt` trait allows access to other methods besides
+//! `from_args`. You can, for instance, get the underlying `clap::ArgMatches`
+//! by doing `Opt::clap().get_matches()`. See the [trait's reference documentation](trait.StructOpt.html)
+//! for more details.
```
CreepySkeleton: You should use `[trait's reference documentation](StructOpt)` instead of explicit file name, see [the RFC](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md)

thiagoarrais: Can't get this to work. I've tried these two syntaxes:

```md
    [trait's reference documentation](StructOpt)
```

```md
    [trait's reference documentation](structopt::StructOpt)
```

I'm running `cargo doc` and rustdoc seems to be version 1.34.2. Should I do anything differently?

CreepySkeleton: I can't make it work either :cry: Forget about it.

TeXitoi:
Thanks!

thiagoarrais:
No problem, Texitoi. I'm very glad to contribute to structopt!

Issue #265, opened 2019-09-20
Unnecessary name sanitisation


Fiedzia:
Clap has no problem with subcommand names starting with colon.
I need this to namespace them (users will be able to create their own, so I want standard ones not to clash witht them.

However something in structopt removes them (and anything that's not a letter) from beginning and end of subcommand name. This seems accidental.

CreepySkeleton:
It seems like influence of `rename_all`.

TeXitoi:
That's a feature, but you can use
```rust
#[structopt(rename_all = "verbatim")]
```
to manage the names by hand.

CreepySkeleton:
@TeXitoi That's right, but I never thought that dots `.` or other punctuation marks should be renamed by this, `rename_all` meant to rename `SomeCase` to `some-case` and the like?

CreepySkeleton:
Err, wait, there's no way to start a name with a colon except for explicit `name = ":asda"`. @Fiedzia do you already do this?

Fiedzia:
I am explicitely setting name with name=":somevalue", that doesn't work.
rename_all = "verbatim" does work.

This behavior was surprising though, especially that using clap directly doesn't do that.

CreepySkeleton:
> but it's technically a breaking change.

Technically - yes, it is, but I highly doubt that much user has been relying on this change. We should really ask on `users.rust-lang.org` or `reddit`

CreepySkeleton:
@TeXitoi I think this is basically a bug, and semver states that bugfixes does not require major version bump.

I'd like to conduct a survay on users.rust-lang.org, do you mind?

TeXitoi:
I'm OK with doing it even without asking, but feel free to ask if you prefer to be sure.

CreepySkeleton:
@TeXitoi this is closed too

Issue #269, opened 2019-10-04
Raw ident for struct member prefixes "r-" to parameter and value names


Arnavion:
```rust
#[derive(structopt::StructOpt)]
struct Options {
    #[structopt(long)]
    r#type: String,
}

fn main() {
    let _: Options = structopt::StructOpt::from_args();
}
```

```
$ cargo run -- --help

rust-test 0.1.0

USAGE:
    rust-test --r-type <r-type>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --r-type <r-type>
```

The parameter long name `--r-type` should be `--type`, and the parameter value string `r-type` should be `type`. Basically the `r#` of the struct member ident should not contribute to the `clap::Arg` parameters.

(Can be worked around by setting `#[structopt(long = "type", name = "type")]` explicitly.)

CreepySkeleton:
@TeXitoi This is more of a bug than enhancement

TeXitoi:
I put it as enhancement as this case was not documented, and the result, even if surprising, is not problematic. I agree that we can correct this without.

Arnavion:
Technically it would be a breaking change for anyone who was relying on the behavior, eg their scripts (and the scripts of their binary's users) that used `--r-type` would stop working. Though of course that's probably not very likely.

PR #277, opened 2019-10-29
Clarify docs and error messages

CreepySkeleton:
Docs are never clear enough, perfection is a moving goal but it doesn't
mean we should stop :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -104,10 +104,13 @@
 //! (unless they're [subcommands](#subcommands)),
 //! and you can adjust these apps and args by `#[structopt(...)]` [attributes](#attributes).
 //!
-//! The derived `StructOpt` trait allows access to other methods besides
-//! `from_args`. You can, for instance, get the underlying `clap::ArgMatches`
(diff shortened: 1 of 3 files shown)


CreepySkeleton:
@TeXitoi what do you think about https://github.com/clap-rs/clap_derive/pull/22 ? Do you agree to deprecate `structopt` once `clap 3` is out? I'll fix #265 and #269 within a day and then I'm going to finish the transition, if you don't mind.

TeXitoi:
That's OK for me, but will clap v3 be released within a short period?

I propose that structopt stay on clap v2, with a maintenance only mode when clap v3 with featureful clap derive is out.

CreepySkeleton:
@Dylan-DPC @kbknapp What do you think is anticipated release date for `clap v3`?



> I propose that structopt stay on clap v2, with a maintenance only mode when clap v3 with featureful clap derive is out.

Yeah, that's what I meant by "deprecate". I just think that you should add a big note somewhere in the README.md about it.

PR #278, opened 2019-11-04
Fix #269 and #265

### LINKED ISSUE #269: Raw ident for struct member prefixes "r-" to parameter and value names

CreepySkeleton:
Closes #269 #265 
Single PR because this changes are on top of one another

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -77,9 +77,14 @@ pub enum CasingStyle {
 }
 
 #[derive(Clone)]
+pub enum Name {
+    Derived(Ident),
+    Assigned(LitStr),
+}
+
+#[derive(Clone)]
 pub struct Attrs {
-    name: Sp<String>,
-    cased_name: String,
+    name: Name,
     casing: Sp<CasingStyle>,
     methods: Vec<Method>,
     parser: Sp<Parser>,
@@ -148,7 +153,9 @@ impl Parser {
 
         let func = match spec.parse_func {
             None => match kind {
-                FromStr | FromOsStr => quote_spanned!(spec.kind.span()=> ::std::convert::From::from),
+                FromStr | FromOsStr => {
+                    quote_spanned!(spec.kind.span()=> ::std::convert::From::from)
+                }
                 TryFromStr => quote_spanned!(spec.kind.span()=> ::std::str::FromStr::from_str),
                 TryFromOsStr => span_error!(
                     spec.kind.span(),
@@ -171,19 +178,6 @@ impl Parser {
 }
 
 impl CasingStyle {
-    fn translate(&self, input: &str) -> String {
-        use CasingStyle::*;
-
-        match self {
-            Pascal => input.to_camel_case(),
-            Kebab => input.to_kebab_case(),
-            Camel => input.to_mixed_case(),
-            ScreamingSnake => input.to_shouty_snake_case(),
-            Snake => input.to_snake_case(),
-            Verbatim => String::from(input),
-        }
-    }
-
     fn from_lit(name: LitStr) -> Sp<Self> {
         use CasingStyle::*;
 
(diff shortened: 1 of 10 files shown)


structopt-derive/src/attrs.rs
```diff
@@ -76,10 +77,17 @@ pub enum CasingStyle {
     Verbatim,
 }
 
+#[derive(Clone)]
+pub enum Name {
+    Derived(Ident),
+    Assigned(LitStr),
+}
```
TeXitoi: Can't the name be converted upfront instead at usage? It would remove the need of this enum and might simplify the code.

structopt-derive/src/attrs.rs
```diff
@@ -541,7 +560,7 @@ impl Attrs {
                     }
                     Ty::OptionVec => {
                         // If it's a positional argument.
-                        if !(res.has_method("long") || res.has_method("short")) {
+                        if !(res.long.is_some() || res.short.is_some()) {
```
TeXitoi: Should use is_positional?

structopt-derive/src/parse.rs
```diff
@@ -28,13 +28,13 @@ impl Parse for StructOptAttributes {
...
     // single-identifier attributes
-    Short(Ident),
-    Long(Ident),
     Flatten(Ident),
     Subcommand(Ident),
     NoVersion(Ident),
 
     // ident [= "string literal"]
+    Short(Ident, Option<LitStr>),
+    Long(Ident, Option<LitStr>),
```
TeXitoi: Converting upfront would not need this change

structopt-derive/src/spanned.rs
```diff
@@ -85,8 +89,13 @@ impl<T: AsRef<str>> AsRef<str> for Sp<T> {
...
     fn to_tokens(&self, stream: &mut TokenStream) {
-        let val = &self.val;
-        let quoted = quote_spanned!(self.span=> #val);
-        stream.extend(quoted);
+        // this is the simplest way out of correct ones to change span on
+        // arbitrary token tree I can come up with
+        let tt = self.val.to_token_stream().into_iter().map(|mut tt| {
+            tt.set_span(self.span.clone());
+            tt
+        });
```
TeXitoi: I don't get this modification, can you explain?

structopt-derive/src/attrs.rs
CreepySkeleton: Yeah

structopt-derive/src/attrs.rs
CreepySkeleton: Why do you so strongly advocate for an upfront conversion? Actually, I *have* tried applying a "simple" fix but it turned out that the code would be *much* more complex than it is now. Also, I would argue that such an enum represents the intention just perfectly.

structopt-derive/src/parse.rs
CreepySkeleton: In this case `ling = "..."` would count as a raw method but we still need to apply `unraw` to it. If you think searching for it in raw methods and applying the fix after-the-fact would be simpler/better - well, you better think again :)

structopt-derive/src/spanned.rs
CreepySkeleton: Previous approach was to use `quote_spanned!(self.span => #val)` but it didn't really work. `quote_spanned!` can't really apply this span to interpolated tokens so it used `#val.span()` (pseudocode). In most cases this span turns out to be `Span::call_site`. Current approach is 100% correct and doesn't really change anything.

CreepySkeleton:
I'll be adding fixes as new commits, once you're happy with the PR I'll rebase to get clean 3-commpi PR

structopt-derive/src/attrs.rs
CreepySkeleton: Hmm, perhaps we don't really need `Name` for `long/short` but it's still necessary for `name`

CreepySkeleton:
OK, @TeXitoi , now it's as simple as I can manage, tell me if you have any more ideas. Also, just released `rustc 1.39` changed error message format a bit so I set it to ignore ui tests on pre-1.39, is that OK?

TeXitoi:
Great, thanks!

TeXitoi:
v0.3.4 published.

TeXitoi:
@CreepySkeleton Do you want to be official contributor of structopt? If yes, I give you commit access to this repository. No pressure, just I will make PR and you'll have to read and merge them, but no other engagements.

CreepySkeleton:
@TeXitoi I agree with pleasure :smile: Just be aware that I can't guarantee any determinate involvement/ping time. Normally, I should be able to reply in ~5 hours but life get complicated sometimes.

TeXitoi:
No pressure.

PR #191, opened 2019-05-30
Support Option<Vec<T>> field type

sphynx:
Closes #180 

The code is quite similar to adding support of `Option<Option<T>>`

PR #279, opened 2019-11-05
Fix parse for OptionVec ref pull #191

xiaoniu-578fa6bff964d005:
Old implementation will forbid parse for Option<Vec<T>> and the error message is not helpful.


TeXitoi:
v0.3.4 published.

TeXitoi:
And thanks for the contribution.

PR #282, opened 2019-11-22
Pass the try_from_str functions a &str instead of a &String.

m-ou-se:
In most cases this doesn't matter, as &String is coerced to a &str, but
this fails for generic functions like CString::new.

The added test failed to compile before this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Upcoming
+
+* `try_from_str` functions are now called with a `&str` instead of a `&String` ([#282](https://github.com/TeXitoi/structopt/pull/282))
+
 # v0.3.4 (2019-11-08)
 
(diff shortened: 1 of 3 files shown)


TeXitoi:
Thanks for your contribution!

Some questions:
 - Don't we have the same problem with `from_str`, `from_ostr` and `try_from_os_str`
 - Don't we need to do this fix at some other places to be safe

Please also add a line in the CHANGELOG.md

m-ou-se:
It was only a problem in the validation step, not in the conversion step. So the non-`try_` variants are fine.

`try_from_os_str` is also fine, because Clap's `validator_os` gives an `&OsStr` to the function. Only Clap's `validator` is a bit weird, as it gives a `String` (by value).

TeXitoi:
Thanks.

TeXitoi:
v0.3.5 published

PR #284, opened 2019-11-26
Fix #283

### LINKED ISSUE #283: `version = expr` ignored, and incorrectly uses `env!(CARGO_PKG_VERSION)`
Setting `#[structopt(version = expr)]` (where `expr` is any expression, but not a literal) the expression is ignored and `env!(CARGO_PKG_VERSION)` is used instead.

It looks like the problem is `clap::App::version` is being called twice, first time with `expr`, and second time with the value of `CARGO_PKG_VERSION`, and the second call wins. In the expanded macro output I'm seeing something like:

``` rust
let app = ::structopt::clap::App::new("example")
    .version(VERSION)
    .version("0.1.0");
```

Here is a reduction of my code that's seeing the problem:

``` rust
use structopt::StructOpt;

const VERSION: &str = "custom version";

#[derive(StructOpt)]
#[structopt(version = VERSION)]
pub struct Opts {}

fn main() {
    let _ = Opts::from_args();
}
```

As a workaround I can set `#[structopt(no_version, version = VERSION)]` and get the result I want, although looking at [structopt_derive::attrs::Attrs::top_level_methods][tlm] it doesn't look like this was intended to work.

[tlm]: https://github.com/TeXitoi/structopt/blob/b5907e8996d39d2bcecd5aeb04466a256611eb9e/structopt-derive/src/attrs.rs#L591

CreepySkeleton:
Closes #283

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -604,7 +604,7 @@ impl Attrs {
         let about = &self.about;
         let methods = &self.methods;
 
-        quote!( #(#methods)* #author #about #version )
+        quote!( #author #about #version #(#methods)* )
     }
 
     /// generate methods on top of a field
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -604,7 +604,7 @@ impl Attrs {
         let about = &self.about;
         let methods = &self.methods;
 
-        quote!( #author #about #version #(#methods)* )
+        quote!( #author #version #(#methods)* #about )
     }
 
     /// generate methods on top of a field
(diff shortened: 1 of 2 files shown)


TeXitoi:
Better to not merge himself without a review ;-)

CreepySkeleton:
Well, this was a small straightforward change, covered by a test, so I decided not to bother you. I would certainly wait for you approval otherwise. Do you require to ask for you approval every time? (I'm OK with that, I just meant to take some of the load off your shoulders)

TeXitoi:
I think that's always better to have an external view. Same for me. We can say that if we don't have feedback within a day and that you don't need feedback, we can merge ourself?

Issue #285, opened 2019-11-26
v0.3 support for Option<Vec<T>>


ryankurte:
hey thanks for making a super useful crate!

i've just updated from v0.2.x to v0.3.x and am running into a problem with parsing vectors that i'm not sure how to resolve.

In `v0.2.x` the following is valid, and passing an option `-d something` ends up with a vector containing `something`.
```rust
#[derive(Debug, Clone, PartialEq, StructOpt)]
pub struct Options {
    #[structopt(short = "d", long = "data", parse(from_str = "data_from_str"))]
    /// Data as a string
    pub data: Option<Vec<u8>>,
}

pub fn data_from_str(s: &str) -> Vec<u8> {
    s.as_bytes().to_vec()
}
```

In `v0.3.x` this results in an error from structopt derive
```rust
#[derive(Debug, Clone, PartialEq, StructOpt)]
pub struct Options {
    #[structopt(short = "d", long = "data", parse(from_str = data_from_str))]
    /// Data as a string
    pub data: Option<Vec<u8>>,
}

pub fn data_from_str(s: &str) -> Vec<u8> {
    s.as_bytes().to_vec()
}
```

```
error[E0308]: mismatched types
 --> src/main.rs:8:15
  |
8 |     pub data: Option<Vec<u8>>,
  |               ^^^^^^ expected u8, found struct `std::vec::Vec`
  |
  = note: expected type `std::vec::Vec<u8>`
             found type `std::vec::Vec<std::vec::Vec<u8>>`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: could not compile `current`.

To learn more, run the command again with --verbose.
```

is it still possible to achieve this in `v0.3.x`?

CreepySkeleton:
This is because `Option<Vec<T>>` [is a special case now](https://github.com/TeXitoi/structopt/blob/master/CHANGELOG.md#support-optional-vectors-of-arguments-for-distinguishing-between--o-1-2--o-and-no-option-provided-at-all-by-sphynx-180), it's different from `Option<T>`.

To suppress this special casing you can use a type alias, like described in changelog
```rust
type Something = Vec<u8>;

#[derive(StructOpt)]
struct Opt {
    #[structopt(short = "d", long = "data", parse(from_str = data_from_str))]
    /// Data as a string
    pub data: Option<Something>,
}
```

[...]

PR #287, opened 2019-11-28
Handle special types correctly

CreepySkeleton:
While thinking on #285 I suggested to use `::vec::Vec` instead of `Vec` to prevent special casing, but a simple test proved me wrong. In fact, **any path that ends with `Vec<T>` will be considered as `Vec<T>`** which is wrong. The same goes for `Option` and `bool`.

I believe this is a bug and it should be fixed. This PR handles paths correctly, checking not only the last segment but also verifying that the said segment is *the only one* segment present, no leading `::`.

Technically, this is a breaking change but I believe this is a bug and, furthermore, I really doubt someone out there has been relying on this behavior.

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use crate::{parse::*, spanned::Sp};
+use crate::{parse::*, spanned::Sp, ty::Ty};
 
 use std::env;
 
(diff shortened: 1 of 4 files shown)


TeXitoi:
There is some cases when it could be used: to have something collectable that is not a vec, one can declare something that ends with Vec to trick structopt.

I don't think neither that someone rely on that, so let's merge it. At worst, we will revert it if we have some bug reports.

TeXitoi:
We forgot the changelog for this one.

CreepySkeleton:
I'll remember to make sure it'll end up being put in changelog one way or another, don't you worry :)

PR #288, opened 2019-11-29
Cleanup tests

CreepySkeleton:


tests/deny-warnings.rs
```diff
@@ -7,12 +7,12 @@
...
-#![feature(never_type)]
 
 use structopt::StructOpt;
 
-fn try_str(s: &str) -> Result<String, !> {
+// FIXME(@CreepySkeleton):
+// replace the return type with `Result<String, !>` once "never type" is stable
+fn try_str(s: &str) -> Result<String, i32> {
     Ok(s.into())
 }
```
CreepySkeleton: I'm asking for a review because of this change. `nightly` feature was removed few months ago so this tests have been ignored since then. Just so you're aware

TeXitoi: Better to return `Result<String, std::convert::Infallible>`, it will do the migration automatically.

PR #291, opened 2019-12-02
cargo clippy

CreepySkeleton:


structopt-derive/src/parse.rs
```diff
@@ -18,11 +18,10 @@ pub struct StructOptAttributes {
...
     fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
         let content;
+        let paren_token = parenthesized!(content in input);
+        let attrs = content.parse_terminated(StructOptAttr::parse)?;
 
-        Ok(StructOptAttributes {
-            paren_token: parenthesized!(content in input),
-            attrs: content.parse_terminated(StructOptAttr::parse)?,
-        })
+        Ok(StructOptAttributes { paren_token, attrs })
```
TeXitoi: Why clippy doesn't like that?

CreepySkeleton: Because initialization order is not really defined (yet). De-facto it is left-to-right but relying on this is relying on unspecified behavior.

TeXitoi: OK, thanks, good to know, I use that sometime.

CreepySkeleton: I didn't know about it either until clippy told me :) The more you know and all...

PR #292, opened 2019-12-02
Add true boolean argument to example

### LINKED ISSUE #212: Weird bool behaviour when not defined as a flag parameter
I've been (incorrectly) using `bool` arguments without any option under the assumption that the mandatory value is actually parsed and turned into a proper `bool` according to what was passed on the CLI but I just found out that it is always `true`. This makes `bool` parameters which are not flags rather useless; IMHO they either need to be really parsed or there should be a compiler error saying: `You've been trying to use a bool parameter which is not a flag and hence will be always true after evaluation which is probably not what you want`.

therealprof:
Regular bool types are only flags and their parsed value represents whether they were present on the command line or absent. However it is very useful to have true boolean argument parsing  which can be combined with `Option` or default values.

Closes #212

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,9 @@ And then, in your rust file:
 use std::path::PathBuf;
 use structopt::StructOpt;
 
+/// Special type to be used for `true` and `false` parsing
+type Bool = bool;
+
 /// A basic example
 #[derive(StructOpt, Debug)]
 #[structopt(name = "basic")]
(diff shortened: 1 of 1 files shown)


CreepySkeleton:
Again, I don't think that such a corner case should be on the front page, this example is simple for a reason. 

Would you be happy with [this](https://github.com/TeXitoi/structopt/pull/293/files) variant?

PR #293, opened 2019-12-02
Prohibit positional bool

### LINKED ISSUE #212: Weird bool behaviour when not defined as a flag parameter

CreepySkeleton:
Closes #212

cc @TeXitoi please also take a look at #292

### DIFF
--- a/examples/true_or_false.rs
+++ b/examples/true_or_false.rs
@@ -1,0 +1,41 @@
+//! How to parse `--foo=true --bar=false` and turn them into bool.
+
+use structopt::StructOpt;
+
+fn true_or_false(s: &str) -> Result<bool, &'static str> {
+    match s {
+        "true" => Ok(true),
+        "false" => Ok(false),
+        _ => Err("expected `true` or `false`"),
+    }
+}
+
+#[derive(StructOpt, Debug, PartialEq)]
+struct Opt {
+    // Default parser for `try_from_str` is FromStr::from_str.
+    // `impl FromStr for bool` parses `true` or `false` so this
+    // works as expected.
+    #[structopt(long, parse(try_from_str))]
+    foo: bool,
+
+    // Of course, this could be done with an explicit parser function.
+    #[structopt(long, parse(try_from_str = true_or_false))]
+    bar: bool,
+
+    // `bool` can be positional only with explicit `parse(...)` annotation
+    #[structopt(long, parse(try_from_str))]
+    boom: bool,
+}
+
+fn main() {
+    assert_eq!(
+        Opt::from_iter(&["test", "--foo=true", "--bar=false", "true"]),
+        Opt {
+            foo: true,
+            bar: false,
+            boom: true
+        }
+    );
+    // no beauty, only truth and falseness
+    assert!(Opt::from_iter_safe(&["test", "--foo=beauty"]).is_err());
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -302,6 +302,22 @@
 //! If you would like to use a custom string parser other than `FromStr`, see
 //! the [same titled section](#custom-string-parsers) below.
 //!
+//! **Note:**
+//! _________________
+//! Pay attention that *only literal occurrence* of this types is special, for example
+//! `Option<T>` is special while `::std::option::Option<T>` is not.
+//!
+//! If you need to avoid special casing you can make a `type` alias and
+//! use it in place of the said type.
+//! _________________
+//!
+//! **Note:**
+//! _________________
+//! `bool` cannot be used as positional argument unless you provide an explicit parser.
+//! If you need a positional bool, for example to parse `true` or `false`, you must
+//! annotate the field with explicit [`#[structopt(parse(...))]`](#custom-string-parsers).
+//! _________________
+//!
 //! Thus, the `speed` argument is generated as:
 //!
 //! ```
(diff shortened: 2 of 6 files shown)


therealprof:
The example is good, the "documentation" does not help a bit.

CreepySkeleton:
@therealprof This to address #287 , your case is covered by the example.

therealprof:
Well, it didn't say that on the label, maybe you should throw in a "Closes #287" then?

CreepySkeleton:
@therealprof 
#287 is not an issue but PR, take look at the first comment there for detailed information.
I included it in this PR because "see this for example" link refers to `examples/true_or_false.rs` which is to be added by this PR.

CreepySkeleton:
> The example is good, the "documentation" does not help a bit.

Also, @therealprof , if you have some idea about how to address your case in the doc I'm all ears.

CreepySkeleton:
@therealprof 

I would like to apologize for being too blunt. I'm not really a "public relations" man, I take strait approach when explaining thinks to others, too strait sometimes. It was never my intention to offend you or discard your opinion, sorry if I left wrong impression.

But I'm not making amends, I mean every word I said. I still believe you case is a corner one, `true/false` arguments are common for config files indeed, but I have yet to meet a case when I would need to actually use them on command line. Even more, I wasn't even aware of existence of such applications until after you'd brought some examples. 

We can't bring every corner case to README (and it's not an actual documentation anyway), hence I'm closing your PR. Do you think this PR provides adequate explanation?

examples/true_or_false.rs
```diff
@@ -0,0 +1,36 @@
+//! How to parse `--foo=true --bar=false` and turn them into bool.
+
+use structopt::StructOpt;
+
+fn true_or_false(s: &str) -> Result<bool, &'static str> {
+    match s {
+        "true" => Ok(true),
+        "false" => Ok(false),
```
TeXitoi: Maybe "yes" or "no" would be a more meaningful example.

examples/true_or_false.rs
```diff
@@ -0,0 +1,36 @@
...
+    // Default parser for `try_from_str` is FromStr::from_str.
+    // `impl FromStr for bool` parses `true` or `false` so this
+    // works as expected.
+    #[structopt(long, parse(try_from_str))]
+    foo: bool,
+
+    // Of course, this could be done with an explicit parser function.
+    #[structopt(long, parse(try_from_str = true_or_false))]
+    bar: bool,
+}
```
TeXitoi: It could be great to also add a boolean parameter for completeness.

src/lib.rs
```diff
@@ -302,6 +302,22 @@
...
+//! Pay attention that *only literal occurrence* of this types is special, for example
+//! `Option<T>` is special while `::std::option::Option<T>` is not.
+//!
+//! If you need to avoid special casing you can make a `type` alias and
+//! use it in place of the said type.
+//! _________________
+//!
+//! **Note:**
+//! _________________
+//! `bool` cannot be used as positional argument with an implicit parser. If you need
```
TeXitoi: I'd prefer "without an explicit parser"

src/lib.rs
```diff
@@ -302,6 +302,22 @@
...
+//! a positional `bool`, for example to parse `true` or `false` add an explicit
+//! [`#[structopt(parse(...))]` annotation](#custom-string-parsers).
```
TeXitoi: If you need a positional `bool`, for example to parse `true` or `false`, you can add an explicit [`#[structopt(parse(try_from_str))]` annotation](#custom-string-parsers).

examples/true_or_false.rs
CreepySkeleton: Do you mean positional parameter?

examples/true_or_false.rs
CreepySkeleton: This example is meant to demonstrate how `impl FromStr for bool` works, the intention is to show how to do the same thing in two different ways.

examples/true_or_false.rs
TeXitoi: yep

examples/true_or_false.rs
TeXitoi: OK

therealprof:
@CreepySkeleton No worries, all good here.

examples/true_or_false.rs
```diff
@@ -0,0 +1,41 @@
...
+        Opt::from_iter(&["test", "--foo=true", "--bar=false", "true"]),
+        Opt {
+            foo: true,
+            bar: false,
+            boom: true
+        }
+    );
+    // no beauty, only truth and falseness
+    assert!(Opt::from_iter_safe(&["test", "--foo=beauty"]).is_err());
+}
```
TeXitoi: Examples are normally working programs that output the Opt struct to stdout, not unit test like.

CreepySkeleton: Yes, I just wanted to show "what's the result is" in the most demonstrative way.

PR #295, opened 2019-12-03
Update proc-macro-error to v0.4

CreepySkeleton:
cc @TeXitoi please double check error messages

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -12,7 +12,7 @@ use std::env;
 
 use heck::{CamelCase, KebabCase, MixedCase, ShoutySnakeCase, SnakeCase};
 use proc_macro2::{Span, TokenStream};
-use proc_macro_error::span_error;
+use proc_macro_error::abort;
 use quote::{quote, quote_spanned, ToTokens};
 use syn::{self, ext::IdentExt, spanned::Spanned, Attribute, Expr, Ident, LitStr, MetaNameValue};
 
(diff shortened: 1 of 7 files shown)


structopt-derive/src/attrs.rs
```diff
@@ -95,8 +95,10 @@ impl Method {
             None => match env::var(env_var) {
                 Ok(val) => LitStr::new(&val, ident.span()),
                 Err(_) => {
-                    span_error!(ident.span(),
-                        "`{}` environment variable is not defined, use `{} = \"{}\"` to set it manually", env_var, ident, ident
+                    abort!(ident.span(),
+                        "cannot derive `{}` from Cargo.toml", ident;
+                        note = "`{}` environment variable is not set", env_var;
+                        help = "use `{} = \"{}\"` to set {} manually", ident, ident, ident;
```
TeXitoi: use `#[structopt({} = \"{}\")]` to set {} manually

structopt-derive/src/parse.rs
```diff
@@ -134,25 +139,35 @@ impl Parse for StructOptAttr {
...
-                        Err(_) => span_error!(name.span(),
-                            "`#[structopt(raw(...))` attributes are removed in structopt 0.3, they are replaced with raw methods\nhelp: if you meant to call `clap::Arg::raw()` method you should use bool literal, like `raw(true)` or `raw(false)`")
+                    Err(_) => {
+                        abort!(name.span(),
+                            "`#[structopt(raw(...))` attributes are removed in structopt 0.3, \
+                            they are replaced with raw methods";
+                            help = "if you meant to call `clap::Arg::raw()` method \
+                                you should use bool literal, like `raw(true)` or `raw(false)`";
+                            note = "if you need to call some method from `clap::Arg/App` you \
+                                should use raw method";
```
TeXitoi: I think "raw method" is not explicit enough, but I don't have a better idea.

structopt-derive/src/attrs.rs
CreepySkeleton: That's right. It'll be turned into "use `about = "about"`  to set about manually. Too much "about"?

structopt-derive/src/parse.rs
CreepySkeleton: Well, we could try to parse erroneous `name = "litstr"` expression and suggest "you should use `name(litstr)`". I would involve pretty complicated post-processing (probably extracted into an external fn) but I believe it would worth the candle, don't you think?

CreepySkeleton: I decided to put some effort to this message since it's not the first time someone came complaining about it being unclear.

CreepySkeleton:
@TeXitoi You can safely delete branches if you merged corresponding PR and the PR is mine 😄

PR #300, opened 2019-12-05
Send headers to all web sites

CreepySkeleton:


CreepySkeleton:
What the hell? I almost feel like writing my own checker just to get rid of `markdown-link-check`

CreepySkeleton:
It supposed to allow regexps in `"urls" = [...]` but it doesn't work, I've tried multiple variants. I'm giving up and going to write my own checker in python. @TeXitoi is it acceptable?

TeXitoi:
Please no. Just having a list will be enough as we will not do a lot of external links.

PR #302, opened 2019-12-05
Introduce #[structopt(rename_all_env)]

### LINKED ISSUE #149: rename_all for env vars
(I'm not sure if this should be on `structopt` or `clap-rs`), 

This is a feature request so `env` may be used without any value (lik short/long), and be automatically (but differently) renamed.

Quick example:

from this:
```rust
#[derive(Clone, Debug, Deserialize, Serialize, StructOpt)]
#[serde(rename_all = "kebab-case")]
#[structopt(rename_all = "kebab-case")]
pub struct Config {
    #[structopt(long, env = "BIND_ADDR", default_value = "127.0.0.1:8080")]
        /// Address which the server will bind to.
        pub bind_addr: SocketAddr,
}
```

into this:
```rust
#[derive(Clone, Debug, Deserialize, Serialize, StructOpt)]
#[serde(rename_all = "kebab-case")]
#[structopt(rename_all = "kebab-case")]
#[structopt(rename_all_env = "SCREAMING_SNAKE_CASE")]
pub struct Config {
    #[structopt(long, env, default_value = "127.0.0.1:8080")]
        /// Address which the server will bind to.
        pub bind_addr: SocketAddr,
}
```

CreepySkeleton:
Closes #149

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -526,7 +526,7 @@
 //! value the value will be derived from the field's name. This is controlled by
 //! `#[structopt(rename_all_env)]`.
 //!
-//! `rename_all_env` works almost exactly as `rename_all` (including overriding)
+//! `rename_all_env` works exactly as `rename_all` (including overriding)
 //! except default casing is `SCREAMING_SNAKE_CASE` instead of `kebab-case`.
 //!
 //! ## Skipping fields
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -475,7 +475,7 @@
 //! #[derive(StructOpt)]
 //! struct Foo {
 //!   #[structopt(short, long, env = "PARAMETER_VALUE")]
-//!   param: String
+//!   parameter_value: String
 //! }
 //! # fn main() {}
 //! ```
@@ -487,7 +487,7 @@
 //! $ cargo run -- --help
 //! ...
 //! OPTIONS:
-//!   -p, --param <param>     [env: PARAMETER_VALUE=env_value]
+//!   -p, --parameter-value <parameter-value>     [env: PARAMETER_VALUE=env_value]
 //! ```
 //!
 //! In some cases this may be undesirable, for example when being used for passing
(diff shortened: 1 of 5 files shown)


CreepySkeleton:
cc @TeXitoi ready for review

src/lib.rs
```diff
@@ -499,11 +499,36 @@
...
+//!   secret_value: String
 //! }
-//! # fn main() {}
 //! ```
 //!
+//! It works just like `#[structopt(short/long)]`: if `env` is not set to some concrete
+//! value the value will be derived from the field's name. This is controlled by
+//! `#[structopt(rename_all_env)]`.
+//!
+//! `rename_all_env` works almost exactly as `rename_all` (including overriding)
```
TeXitoi:
```suggestion
//! `rename_all_env` works exactly as `rename_all` (including overriding)
```
Else we might think there is some subtle undocumented differences.

tests/rename_all_env.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[test]
+fn overridable() {
+    #[derive(Debug, PartialEq, StructOpt)]
+    #[structopt(rename_all_env = "kebab")]
+    struct BehaviorModel {
+        #[structopt(env)]
+        be_nice: String,
+
+        #[structopt(rename_all_env = "pascal", env)]
```
TeXitoi: I think in this case, it should be `rename_env`, because there is no "all" in this context.

CreepySkeleton: I don't quite understand. Do you propose rename the attr to `#[structopt(rename_env)]`?

TeXitoi: Yep, to match serde `rename` and `rename_all`. Now a local rename is, I think, not so interesting.

CreepySkeleton: >  to match serde `rename` and `rename_all`

You seem to be misunderstanding this. 

This attr is not meant to match `serde`'s `rename`/`rename_all`, it's meant to match `structopt`'s `rename_all`.

This specific piece of code tests the "override" feature for `rename_env_all` 
```rust
// we want all the derived flags here to be pascal case
#[structopt(rename_all = "pascal")]
// we also want the derived env fallbacks to use kebab case
#[structopt(rename_all_env = "kebab")]
enum Subcommands {
    // the stuff inside inherits parent renaming policy f
    // long, short, and name will be renamed to pascal case
    // env will be renamed to kebab case
    SetOne { /* ... */ },

    // We want this stuff to be different 
    // we *override* the setting, reapplying the same attributes with different values
    #[structopt(rename_all = "screaming snake", rename_all_env = "camel")]
    SetTwo { ... }

    // all the rest is not affected by the reapplying 
    // but it *is* affected by the parent settings
}
```

This is how you match `structopt`'s attrs to `serde`'s (roughly)
```
#[serde(rename)] == #[structopt(name/short/long/env)] 

#[serde(rename_all)] == #[structopt(rename_all/rename_all_env)]
```

CreepySkeleton:
@TeXitoi so what do you think about this?

tests/rename_all_env.rs
TeXitoi: Yeh, you're right.

TeXitoi:
I'm just septic on the advantage of having `rename_all_env` on a field, but you're right, that the good name for it.

PR #307, opened 2019-12-13
impl StructOpt for Box<impl StructOpt>

### LINKED ISSUE #304: Subcommand configuration causes clippy::large_enum_variant to be emitted
I'm using the (very awesome and useful, by the way) pattern of defining subcommands with specific configurations:

```rust
enum Command {
    #[structopt(name = "version")]
    PrintVersion,

    #[structopt(name = "second")]
    DoSomething {
        #[structopt(flatten)]
        config: DoSomethingConfig,
    },

    #[structopt(name = "first")]
    DoSomethingElse {
        #[structopt(flatten)]
        config: DoSomethingElseConfig,
    }
...
}
```
The problem is that with my current project, consisting of about 5 subcommands with varying configuration complexities, clippy emits the clippy::large_enum_variant warning.

I'm fine with silencing the warning since it's a single point which is not mission-critical at all (parsing command-line args). However it got me wondering if it would be possible to use Boxes to contain the flattened configurations as suggested by the lint... Sounds to me like something that could be supported with Structopt's trait implementation?

Sorry in advance if this is spam, and thanks for this amazing project!

CreepySkeleton:
Closes #304 

This PR makes `is_subcommand`, `from_subcommand`, and `augment_clap` part of the `StructOpt` trait (marking them as `doc(hidden)`. I believe this is justifiable change because:

* The change is 100% backward compatible.
* These methods won't ever collide with user's methods. 
* This way we could add more `impl`s in future with ease.
* The change is fairly small and straightforward.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -812,4 +812,51 @@ pub trait StructOpt {
     {
         Ok(Self::from_clap(&Self::clap().get_matches_from_safe(iter)?))
     }
+
+    // All the following is NOT PUBLIC API!!!
+    //
+    // ** SUBJECT TO CHANGE WITHOUT NOTICE!!! **
+
+    #[doc(hidden)]
+    fn is_subcommand() -> bool {
+        unimplemented!()
+    }
+
+    #[doc(hidden)]
+    fn augment_clap<'a, 'b>(_app: clap::App<'a, 'b>) -> clap::App<'a, 'b> {
+        unimplemented!()
+    }
+
+    #[doc(hidden)]
+    fn from_subcommand<'a, 'b>(_sub: (&'b str, Option<&'b clap::ArgMatches<'a>>)) -> Option<Self>
+    where
+        Self: Sized,
+    {
+        unimplemented!()
+    }
+}
+
+impl<T: StructOpt> StructOpt for Box<T> {
+    fn clap<'a, 'b>() -> clap::App<'a, 'b> {
+        <T as StructOpt>::clap()
+    }
+
+    fn from_clap(matches: &clap::ArgMatches<'_>) -> Self {
+        Box::new(<T as StructOpt>::from_clap(matches))
+    }
+
+    #[doc(hidden)]
+    fn augment_clap<'a, 'b>(app: clap::App<'a, 'b>) -> clap::App<'a, 'b> {
+        <T as StructOpt>::augment_clap(app)
+    }
+
+    #[doc(hidden)]
+    fn is_subcommand() -> bool {
+        <T as StructOpt>::is_subcommand()
+    }
+
+    #[doc(hidden)]
+    fn from_subcommand<'a, 'b>(sub: (&'b str, Option<&'b clap::ArgMatches<'a>>)) -> Option<Self> {
+        <T as StructOpt>::from_subcommand(sub).map(Box::new)
+    }
 }
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -79,7 +79,7 @@ fn gen_augmentation(
 
             let span = field.span();
             let ts = quote! {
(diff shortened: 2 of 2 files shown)


TeXitoi:
The methods should just be declared, not implented with `unimplemented!()`

CreepySkeleton:
The trick here is - what if someone wants not derive but implement `StructOpt` manually? A user would want to implement only `clap` and `from_matches`, they should know nothing about these hidden methods.

TeXitoi:
But if they implement it manually, they expect it to work with generated ones, and that's not the case.

OTOH, adding new interface is a breaking change.

I don't know... I'll think a bit about it, will comeback later.

CreepySkeleton:
Yes, I'm fine with that. It's hard to imagine someone out there was trying to implement `StructOpt` manually anyway :)

TeXitoi:
v0.3.6 published

bkchr:
We implemented StructOpt on our own ^^

bkchr:
https://github.com/paritytech/substrate/blob/master/client/cli/src/params.rs#L957 for reference is someone is interested

CreepySkeleton:
@bkchr OK, I have an idea on how to make this work without adding new methods to the `StructOpt` trait

bkchr:
Don't worry. I just wanted to comment here. I already have a patch for us to fix the problem. I also find it much better to have the method defined in the trait :)

bkchr:
Okay, you already yanked it ^^

CreepySkeleton:
> Okay, you already yanked it ^^

If was yanked for other reasons #315

CreepySkeleton:
Also, @bkchr , why don't you just derive it there?
```rust
#[derive(Clone, Debug, Default, StructOpt)]
#[structopt(no_version)]
pub struct NoCustom {}
```

As far as I can see, it would generate the same code (I mean, the code working the same way) as you have written manually

bkchr:
Good question why I not used `derive(StructOpt)`. I don't really remember. Regarding the `augment_clap` function, you are right. At the time I was writing this code, `StructOpt` seemed to sleep a little bit :) 
I make myself some note to revisit the code and make an issue for the features we need.

Basically we want to support adding extensions to the cli interface, that the users can define.

cecton:
@CreepySkeleton I think that's the missing feature: we would like to be able to extend an enum of subcommands from another crate.

To be clear it doesn't have to use flatten in this fashion. But it would be nice if there would be a way to add a subcommand to an existing list of subcommands. Unfortunately Rust won't allow adding variants in the enum, the usual way to deal with this is this: https://stackoverflow.com/questions/25214064/can-i-extend-an-enum-with-additional-values#25214677 so there should be a way to tell structopt that the subcommands in the variant (`BaseCli::Command1`) should be at the same level than the subcommands in the parent enum (`ExtendedCli::Command2`).

## Code

```
use structopt::StructOpt;

#[derive(Debug, StructOpt)]
enum BaseCli {
    Command1(Command1),
}

#[derive(Debug, StructOpt)]
struct Command1 {
    Argument1: String,
}

#[derive(Debug, StructOpt)]
enum ExtendedCli {
    #[structopt(flatten)]
    BaseCli(BaseCli),
    Command2(Command2),
}

#[derive(Debug, StructOpt)]
struct Command2 {
    Argument2: String,
}

fn main() {
    let cli = ExtendedCli::from_args();
    println!("{:?}", cli);
}
```

## Expected Result

```
test-struct-opt 0.1.0

USAGE:
    test-struct-opt <SUBCOMMAND>

FLAGS:

SUBCOMMANDS:
    command1
    command2
    help        Prints this message or the help of the given subcommand(s)
```

## Actual Result

```
error: flatten is only allowed on fields
  --> src/main.rs:15:17
   |
15 |     #[structopt(flatten)]
   |                 ^^^^^^^
```

PR #308, opened 2019-12-17
chore: Update rustversion to 1

ignatenkobrain:


TeXitoi:
Thanks, but that's already in #296

CreepySkeleton:
@TeXitoi maybe we should reopen this and merge it separately? It's kind of separate feature

TeXitoi:
latest structopt has rustversion 1

PR #312, opened 2019-12-21
Default value

### LINKED ISSUE #123: default_value and Option?
Right now, they are [incompatible](https://github.com/TeXitoi/structopt/blob/master/structopt-derive/src/attrs.rs#L335), but I'm not sure that's right. For example, I wanted something like `my-app --db-path [DB]`, where there's a default database path if not set.

Is there another way to achieve that?

### LINKED ISSUE #142: Feature Request: use`Default`, if implemented
In the same vein as convenience features like #134 and #18, add support for the automatic use of `Default`, if implemented. This would be a breaking change if done poorly, and in any case I'm not sure if the procedural macro system is powerful enough to "detect" trait implementations. I'd therefore propose the following syntax, inspired by #18:

```rust
#[derive(Debug)]
struct Value(String);

impl Default for Value {
    fn default() -> Value {
        Value("default".to_string())
    }
}

impl FromStr for Value { ... }

#[derive(Debug, StructOpt)]
struct Opt {
    // using the proposed automatic name syntax
   #[structopt(long, short, default)]
    value: Value
}
```

Basically, if the `default` key is present, then use the `Default` trait if the flag isn't given. I'd also be okay with `default_value` being the key, but my suggestion is that it be slightly different because under the hood, something slightly different is happening. This isn't using `clap`'s `default_value` builder method, since it requires a str. Rather, `structopt` itself will handle the logic for filling in the default value if necessary.

This could have implications for #123, as well.

### LINKED ISSUE #150: default_value from lazy Default
(Hello, I'm still not sure if this should be on `structopt` or `clap-rs`..), 

This is a feature request so `default_value` may be used without any value (like short/long), and be lazily derived from Default itself.

Quick example:

from this:
```rust
#[derive(Clone, Debug, StructOpt)]
pub struct Config {
    #[structopt(raw(default_value = "&DEFAULT_CONFIG_BIND_ADDR_STR"))]
    pub bind_addr: SocketAddr,
}

impl Default for Config {
    fn default() -> Self {
        Self {
            bind_addr: FromStr::from_str("127.0.0.1:8080").unwrap()
        }
    }
}

// interface between `Default::default()` and (structopt) `default_value`
lazy_static! {
    static ref DEFAULT_CONFIG: Config = Config::default();
    static ref DEFAULT_CONFIG_BIND_ADDR_STR: String = DEFAULT_CONFIG.bind_addr.to_string();
    // (for fields that has no to_string(), I'm "manually" implementing (interfacing) it with some serde serialization by implementing Display)
}
// lazy_static is needed so &'static str may be used. I don't know how different parsing choices would be dealt with..
```

into this:
```rust
#[derive(Clone, Debug, StructOpt)]
#[structopt(default_value)]
pub struct Config {
    pub bind_addr: SocketAddr,
}

impl Default for Config {
    fn default() -> Self {
        Self {
            bind_addr: FromStr::from_str("127.0.0.1:8080").unwrap()
        }
    }
}
```

---

Sorry in advance if this is actually a bad design choice (then I should have no intention of asking for such feature), and that I can't really point into any macro derivation specifics. It's magic !!

CreepySkeleton:
This is not meant to be released with the next patch, just some ongoing work.

Closes #123 
Closes #142
Closes #150 

TODO: docs, changelog

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,6 +41,7 @@
 //! - Arguments
 //!     - [Type magic](#type-magic)
 //!     - [Specifying argument types](#specifying-argument-types)
+//!     - [Default values](#default-values)
 //!     - [Help messages](#help-messages)
 //!     - [Environment variable fallback](#environment-variable-fallback)
 //! - [Skipping fields](#skipping-fields)
@@ -259,7 +260,12 @@
 //!
 //!     Usable only on field-level.
 //!
-//! - [`rename_all`](#specifying-argument-types): [`rename_all = "kebab"/"snake"/"screaming-snake"/"camel"/"pascal"/"verbatim"]`
+//! - [`defautl_value`](#default-values): `default_value [= "default value"]`
+//!
+//!     Usable only on field-level.
+//!
+//! - [`rename_all`](#specifying-argument-types):
+//!     [`rename_all = "kebab"/"snake"/"screaming-snake"/"camel"/"pascal"/"verbatim"]`
 //!
 //!     Usable both on top level and field level.
 //!
@@ -283,7 +289,8 @@
 //!
 //!     Usable only on field-level.
 //!
-//! - [`rename_all_env`](##auto-deriving-environment-variables): [`rename_all_env = "kebab"/"snake"/"screaming-snake"/"camel"/"pascal"/"verbatim"]`
+//! - [`rename_all_env`](##auto-deriving-environment-variables):
+//!     [`rename_all_env = "kebab"/"snake"/"screaming-snake"/"camel"/"pascal"/"verbatim"]`
 //!
 //!     Usable both on top level and field level.
 //!
@@ -315,7 +322,7 @@
 //! If you would like to use a custom string parser other than `FromStr`, see
 //! the [same titled section](#custom-string-parsers) below.
 //!
-//! **Note:**
+//! **Important:**
 //! _________________
 //! Pay attention that *only literal occurrence* of this types is special, for example
 //! `Option<T>` is special while `::std::option::Option<T>` is not.
@@ -418,6 +425,47 @@
 //! #    &["test", "--foo-option", "", "-b", "", "--baz", "", "--custom", "", "positional"]));
 //! # }
 //! ```
+//!
+//! ## Default values
+//!
+//! In clap, default values for options can be specified via [`Arg::default_value`].
+//!
+//! Of course, you can use as a raw method:
+//! ```
+//! # use structopt::StructOpt;
+//! #[derive(StructOpt)]
+//! struct Opt {
+//!     #[structopt(default_value = "", long)]
+//!     prefix: String
+//! }
+//! ```
+//!
+//! This is quite mundane and error-prone to type the `"..."` default by yourself,
+//! especially when the Rust ecosystem uses the [`Default`] trait for that.
+//! It would be wonderful to have `structopt` to take the `Default_default` and fill it
+//! for you. And yes, `structopt` can do that.
+//!
+//! Unfortunately, `default_value` takes `&str` but `Default::default`
+//! gives us some `Self` value. We need to map `Self` to `&str` somehow.
+//!
+//! `structopt` solves this problem via [`ToString`] trait.
+//!
+//! To be able to use auto-default the type must implement *both* `Default` and `ToString`:
+//!
+//! ```
+//! # use structopt::StructOpt;
+//! #[derive(StructOpt)]
+//! struct Opt {
+//!     // just leave the `= "..."` part and structopt will figure it for you
+//!     #[structopt(default_value, long)]
+//!     prefix: String // `String` implements both `Default` and `ToString`
+//! }
+//! ```
+//!
+//! [`Default`]: https://doc.rust-lang.org/std/default/trait.Default.html
+//! [`ToString`]: https://doc.rust-lang.org/std/string/trait.ToString.html
+//! [`Arg::default_value`]: https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.default_value
+//!
 //!
 //! ## Help messages
 //!
@@ -583,8 +631,6 @@
 //! ///
 //! /// Hello!
 //! ```
-//!
-//! Summary
 //! ______________
 //!
 //! [`App::about`]:      https://docs.rs/clap/2/clap/struct.App.html#method.about
@@ -921,8 +967,9 @@ pub use structopt_derive::*;
 
 use std::ffi::OsString;
 
-/// Re-export of clap
+/// Re-exports
 pub use clap;
+pub use lazy_static;
 
 /// A struct that is converted from command line arguments.
 pub trait StructOpt {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -969,6 +969,9 @@ use std::ffi::OsString;
 
 /// Re-exports
 pub use clap;
+
+/// **This is NOT PUBLIC API**.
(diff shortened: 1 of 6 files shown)


CreepySkeleton:
Ready for review

structopt-derive/src/attrs.rs
```diff
@@ -291,6 +296,31 @@ impl Attrs {
...
+                                "#[structopt(default_value)] (without an argument) can be used \
+                                only on field level";
+
+                                note = "see \
+                                    https://docs.rs/structopt/0.3.5/structopt/#magical-methods")
+                        };
+                        quote_spanned!(ident.span()=> {
+                            let val = <#ty as ::std::default::Default>::default();
+                            let s = ::std::string::ToString::to_string(&val);
+                            ::std::boxed::Box::leak(s.into_boxed_str())
```
TeXitoi: What about using a lazy static to have only one allocation by structure (instead of one by call)?

structopt-derive/src/parse.rs
```diff
@@ -39,6 +39,7 @@ pub enum StructOptAttr {
     // ident [= "string literal"]
     About(Ident, Option<LitStr>),
     Author(Ident, Option<LitStr>),
+    DefaultValue(Ident, Option<LitStr>),
```
TeXitoi: Default value can also be a global variable, not necessary a litstr.

TeXitoi: Or litstr name is misleading?

tests/default_value.rs
```diff
@@ -0,0 +1,12 @@
...
+
+#[test]
+fn auto_default_value() {
+    #[derive(StructOpt, PartialEq, Debug)]
+    struct Opt {
+        #[structopt(default_value)]
+        arg: i32,
+    }
+    assert_eq!(Opt { arg: 0 }, Opt::from_iter(&["test"]));
+    assert_eq!(Opt { arg: 1 }, Opt::from_iter(&["test", "1"]));
```
TeXitoi: Please also test that 0 is in the help string.

TeXitoi:
Overall great code!

structopt-derive/src/attrs.rs
CreepySkeleton: I had been thinking about `lazy_static` approach too but I decided against it because I believe that the cons overweight the pros.

Pros:
* Only one leak per `default_value`, usage in loops is safe.

Cons:
* We would need to reexport `lazy_static`, one more dependency, hurts the build time. (Admittedly, not a big concern). 
* This would greatly complicate the code. Current `structopt`'s implementation is nothing but depth-first graph traversing (think about immutable visitors). This structure doesn't play well with popping some info up to the top level; we need all the info on the top because the `lazy_static! { ... }` invocation must be generated just once. 

  I'm not saying this isn't doable, I'm saying this would require a full-scale refactoring. I have one on mind already, but I'd rather postpone it until the release of `clap_derive`, we've been delaying it long enough :)

  Also, this particular problem can be very easily fixed by changing `Arg::default_value` signature to take `Cow<'b str, String>` instead of `'b str`. I believe we should discuss this with the clap-rs team.

Also, I doubt people will complain about this leaks much; even if they will, we can always rework it to using `lazy_static`.

structopt-derive/src/parse.rs
CreepySkeleton: I'm not sure I'm understanding you properly. 

This is just the parser, this code simply parses cases like 
```rust
#[structopt(default_value = "...")]
//                          ^^^^^  <-- this is represented as `LitStr` in syn
```

Cases like this will be handled by `NameLitStr` variant (i.e. just as regular raw method):
```rust
lazy_static! {
     static ref DEFAULT: &str = "...";
}

#[structopt(default_value = DEFAULT)]
//                          ^^^^^  <-- this is represented as `LitStr` in syn
```

By the way, this is exactly how `long`, `short`, `author` and the like work.

structopt-derive/src/attrs.rs
TeXitoi: You can just use a lazy static exactly here without any refactor:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=546d96056c7b65b9eb5198793a40c1d7

structopt-derive/src/parse.rs
TeXitoi: OK, my bad.

CreepySkeleton:
Ready

structopt-derive/src/attrs.rs
```diff
@@ -291,6 +296,39 @@ impl Attrs {
...
+                            abort!(
+                                ident.span(),
+                                "#[structopt(default_value)] (without an argument) can be used \
+                                only on field level";
+
+                                note = "see \
+                                    https://docs.rs/structopt/0.3.5/structopt/#magical-methods")
+                        };
+
+                        let static_name = format_ident!("__STRUCTOPT_DEFAULT_VALUE_{}", fresh_id());
```
TeXitoi: Do you really need this Mame generation? Can't you simply rely on scope for the name of the lazy_static?

structopt-derive/src/attrs.rs
```diff
@@ -291,6 +296,39 @@ impl Attrs {
...
+
+                        quote_spanned!(ident.span()=> {
+                            ::structopt::lazy_static::lazy_static! {
+                                static ref #static_name: &'static str = {
+                                    let val = <#ty as ::std::default::Default>::default();
+                                    let s = ::std::string::ToString::to_string(&val);
+                                    ::std::boxed::Box::leak(s.into_boxed_str())
```
TeXitoi: The leak is not needed, but that's technically the same.

structopt-derive/src/attrs.rs
```diff
@@ -603,6 +642,20 @@ impl Attrs {
...
+
+    thread_local! {
+        static NEXT_ID: Cell<usize> = Cell::new(0);
+    }
+
+    NEXT_ID.with(|next_id| {
+        let id = next_id.get();
+        next_id.set(id + 1);
+        id
+    })
```
TeXitoi: Using an atomic would be simpler and safer if you imagine that there is several threads generating ids. But well, that's just not really important in this case.

structopt-derive/src/attrs.rs
CreepySkeleton: Yeah, no difference, really

structopt-derive/src/attrs.rs
CreepySkeleton: Hmm, I thought the scoping rules are not applicable for statics for some reason. No idea why I thought that, it just ended up in my brain somehow :) I guess I spend too much time with C.

structopt-derive/src/attrs.rs
CreepySkeleton: Good point, but wee don't need it anymore

TeXitoi:
A warning break the error message tests

CreepySkeleton:
Not anymore

src/lib.rs
```diff
@@ -921,8 +967,9 @@ pub use structopt_derive::*;
 
 use std::ffi::OsString;
 
-/// Re-export of clap
+/// Re-exports
 pub use clap;
+pub use lazy_static;
```
TeXitoi: Maybe we can doc(hidden). Or not, as you wish.

CreepySkeleton: The question is whether we want to make the lazy_static public API or not. Let's not.

CreepySkeleton:
Number of issues is less than 15! New achievement! 🚀

PR #313, opened 2019-12-21
Propagate no version

### LINKED ISSUE #242: Inherit no_version for enum variants
In my non-Cargo build environment I end up with a lot of noise from no_version attributes. For subcommands, if the outer enum is no_version I think I shouldn't need to specify no_version as well for each of the command variants.

Currently `derive(StructOpt)` produces an error if any of the no_version attrs here are removed:

```rust
use structopt::StructOpt;

#[derive(StructOpt)]
#[structopt(no_version)]
enum Show {
    #[structopt(no_version)]
    Stats,
    #[structopt(no_version)]
    Capacity,
}
```

CreepySkeleton:
Fixes #242

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# [Upcoming]
+
+* You don't have to apply `#[no_version]` to every `enum` variant anymore.
+  Just annotate the `enum` and the setting will be propagated down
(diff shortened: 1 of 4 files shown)


TeXitoi:
Lack the changelog

PR #314, opened 2019-12-22
Introduce `#[structopt(external_subcommand)]`

### LINKED ISSUE #130: Use of AllowExternalSubcommands
I wonder if it's possible to use [external subcommands](https://docs.rs/clap/2.32.0/clap/enum.AppSettings.html#variant.AllowExternalSubcommands).

I tried adding some `raw(setting = "structopt::clap::AppSettings::AllowExternalSubcommands")` here and there, but `Opt::from_args()` always shows the help then exits.

CreepySkeleton:
Closes #130 

I took some time to develop and debug, but it's finally ready. I have one concern thought.

### DIFF
--- a/tests/subcommands.rs
+++ b/tests/subcommands.rs
@@ -211,3 +211,93 @@ fn flatten_enum() {
         }
     );
 }
+
+#[test]
+fn external_subcommand() {
+    #[derive(Debug, PartialEq, StructOpt)]
+    struct Opt {
+        #[structopt(subcommand)]
+        sub: Subcommands,
+    }
+
+    #[derive(Debug, PartialEq, StructOpt)]
+    enum Subcommands {
+        Add,
+        Remove,
+        #[structopt(external_subcommand)]
+        Other(Vec<String>),
+    }
+
+    assert_eq!(
+        Opt::from_iter(&["test", "add"]),
+        Opt {
+            sub: Subcommands::Add
+        }
+    );
+
+    assert_eq!(
+        Opt::from_iter(&["test", "remove"]),
+        Opt {
+            sub: Subcommands::Remove
+        }
+    );
+
+    assert_eq!(
+        Opt::from_iter(&["test", "git", "status"]),
+        Opt {
+            sub: Subcommands::Other(vec!["git".into(), "status".into()])
+        }
+    );
+
+    assert!(Opt::from_iter_safe(&["test"]).is_err());
+}
+
+#[test]
+fn external_subcommand_os_string() {
+    use std::ffi::OsString;
+
+    #[derive(Debug, PartialEq, StructOpt)]
+    struct Opt {
+        #[structopt(subcommand)]
+        sub: Subcommands,
+    }
+
+    #[derive(Debug, PartialEq, StructOpt)]
+    enum Subcommands {
+        #[structopt(external_subcommand)]
+        Other(Vec<OsString>),
+    }
+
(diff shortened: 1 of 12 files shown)


tests/subcommands.rs
```diff
@@ -211,3 +205,93 @@ fn flatten_enum() {
...
+        Opt::from_iter(&["test", "bar", "--flag", "--arg=0"]),
+        Opt {
+            sub_cmd: SubCmd::Other(vec!["bar".into(), "--flag".into(), "--arg=0".into()])
+        }
+    );
+    assert_eq!(
+        Opt::from_iter(&["test"]),
+        Opt {
+            sub_cmd: SubCmd::Other(vec!["".into()])
+        }
```
CreepySkeleton: What's the behavior should be in this case? We could pass `SubcommandRequiredOrHelp` along with `AllowExternalSubcommands`

TeXitoi: Agree, without option on the subcommand, the subcommand must be provided, and that's not the case here.

CreepySkeleton:
Sorry for the radio silence, the New Year and all :)

This is not abandoned, I'll get to it in 1-2 days.

tests/ui/external_subcommand_wrong_type.stderr
```diff
@@ -0,0 +1,8 @@
+error[E0308]: mismatched types
+  --> $DIR/external_subcommand_wrong_type.rs:13:15
+   |
+13 |     Other(Vec<CString>)
+   |               ^^^^^^^ expected struct `std::ffi::CString`, found struct `std::ffi::OsString`
+   |
+   = note: expected type `std::vec::Vec<std::ffi::CString>`
+              found type `std::vec::Vec<std::ffi::OsString>`
```
CreepySkeleton: This error message is rather unfortunate: I would very much like it to say "expected OsString", but there's nothing we can do about it 😢

CreepySkeleton:
OK, ready to go

tests/ui/external_subcommand_wrong_type.stderr
CreepySkeleton: This is the best I've managed to come up with:

```
error[E0308]: mismatched types
  --> $DIR/external_subcommand_wrong_type.rs:13:15
   |
13 |     Other(Vec<CString>)
   |               ^^^^^^^ expected struct `std::ffi::CString`, found struct `std::ffi::OsString`
   |
   = note: expected type `std::vec::Vec<std::ffi::CString>`
              found type `std::vec::Vec<std::ffi::OsString>`

error[E0308]: mismatched types
  --> $DIR/external_subcommand_wrong_type.rs:13:15
   |
13 |     Other(Vec<CString>)
   |               ^^^^^^^
   |               |
   |               expected struct `std::ffi::OsString`, found struct `std::ffi::CString`
   |               this match expression has type `std::vec::Vec<std::ffi::OsString>`
   |
   = note: expected type `std::vec::Vec<std::ffi::OsString>`
              found type `std::vec::Vec<std::ffi::CString>`

```

While it is indeed possible to draw the error message we desire, I can't get rid of the first one.

CreepySkeleton:
ping @TeXitoi

tests/ui/external_subcommand_wrong_type.stderr
TeXitoi: That's acceptable

PR #316, opened 2019-12-24
V0.3.7

### LINKED ISSUE #315: structopt-derive 0.3.6 breaks structopt <=0.3.5
`structopt-derive v0.3.6` relies on internal details of `structopt v0.3.6`. However, older versions of `structopt` than `v0.3.6` have their dependencies set up so that they can depend on `structopt-derive v0.3.6`. This leads to a compilation failure after someone who has a dependency such as `structopt = "=0.3.5"` runs `cargo update`.

## Steps to reproduce

1. Create a new library package `foo` (`cargo new --lib foo`)
2. Edit `foo/Cargo.toml` to include
  ```toml
  [dependencies]
  structopt = "=0.3.5"
  ```
3. Edit `foo/src/lib.rs` to include
  ```rust
  struct Bar { }
  ```
4. Run `cargo check`
5. You should see a nonsense error, caused by the wrong derive macro being run.

## Possible solution

1. Yank `structopt-derive v0.3.6` and `structopt v0.3.6`
2. In their place release `structopt-derive v 0.4.0` and release an equivalent `structopt v0.3.7` that depends on this new minor version of `structopt-derive`

## Future mitigation
Only release versions of `structopt` that have an `=` dependency on `structopt-derive`. For example if `structopt v0.3.5` had been released with the dependency `structopt-derive = "=0.3.5"` then it wouldn't have been susceptible to this breakage.

CreepySkeleton:
Closes #315 

An attempt to make the hidden methods compatible with cases like [that](https://github.com/paritytech/substrate/blob/master/client/cli/src/params.rs#L957). cc @bkchr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
-# v0.3.6 (2019-12-22)
+# v0.3.7 (2019-12-28)
+
+Nothing's new. Just re-release of `v0.3.6` due to
+[the mess with versioning](https://github.com/TeXitoi/structopt/issues/315#issuecomment-568502792).
+
+You may notice that `structopt-derive` was bumped to `v0.4.0`, that's OK, it's not a breaking change.
+`structopt` will pull the right version in on its on.
+
+# v0.3.6 (2019-12-22) - YANKED
 
 This is unusually big patch release. It contains a number of bugfixes and
 new features, some of them may theoretically be considered breaking. We did our best
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "structopt"
-version = "0.3.6"
+version = "0.3.7"
 edition = "2018"
 authors = ["Guillaume Pinot <texitoi@texitoi.eu>", "others"]
 description = "Parse command line argument by defining a struct."
@@ -28,7 +28,7 @@ travis-ci = { repository = "TeXitoi/structopt" }
 
 [dependencies]
 clap = { version = "2.33", default-features = false }
-structopt-derive = { path = "structopt-derive", version = "0.3.6" }
+structopt-derive = { path = "structopt-derive", version = "=0.4.0" }
 
 [dev-dependencies]
 trybuild = "1.0.5"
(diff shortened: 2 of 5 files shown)


CreepySkeleton:
Let's postpone it until we have come up with a consensus with [this](https://www.reddit.com/r/rust/comments/efi4kb/who_else_depends_on_structopts_internal_api/)

bkchr:
In my opinion it is okay to release with these breaking changes, we will just push a fix to our code base.
However, I would vote for not hiding the functions in the StructOpt trait.

CreepySkeleton:
> release with these breaking changes

Again, none of these changes are breaking in semver sense. If you're relying on internal implementation details - you're mostly on your own.

>  I would vote for not hiding the functions in the StructOpt trait

Those functions are **private** and I'm planning to tweak them a little in #314 . I think we could move them to some sort of `StrutOptMachinery` trait, also hidden.

We *could* stabilize the `augment_clap` method, probably renaming it to `augment_args` since this is what the method does. What do you think?

bkchr:
I just wanted to say, that you should not block because of the code I have written. I will just do the changes required and will come up later with a pr to structopt or at least an issue describing what kind of functionality I need :)

CreepySkeleton:
It's not blocked because of the code someone have written, it's blocked on review from another maintainer, @TeXitoi . If he won't reply in 2 days I'll publish it myself.

CreepySkeleton:
OK, I'm releasing this on my own since @TeXitoi seems to be unavailable right now, and I'm certain I can take the responsibility for this changes. 

`augment_clap` moved to the `StructOptInternal`  trait, also hidden. I'm not stabilizing it yet since nobody showed interest in this topic and, quite frankly, I don't feel that stabilizing anything without an approval from @TeXitoi would be fair.

TeXitoi:
Sorry, didn't find the time to review.

CreepySkeleton:
@TeXitoi you can make it up to me by reviewing #312

TeXitoi:
It would be approved.

PR #323, opened 2020-01-09
CI: build only PRs and pushes to master

CreepySkeleton:
Currently, every PR send by me triggers two identical CI builds: a build for push and a build for PR.

This way we trigger a build only for PRs and pushes to master.

Just stumbled upon it while was surfing across some project :)


CreepySkeleton:
It works!

PR #329, opened 2020-01-15
Flatten on enums

### LINKED ISSUE #327: Adding subcommands to a pre-existing structopt
@CreepySkeleton I think that's the missing feature: we would like to be able to extend an enum of subcommands from another crate.

To be clear it doesn't have to use flatten in this fashion. But it would be nice if there would be a way to add a subcommand to an existing list of subcommands. Unfortunately Rust won't allow adding variants in the enum, the usual way to deal with this is this: https://stackoverflow.com/questions/25214064/can-i-extend-an-enum-with-additional-values#25214677 so there should be a way to tell structopt that the subcommands in the variant (`BaseCli::Command1`) should be at the same level than the subcommands in the parent enum (`ExtendedCli::Command2`).

This issue relates to #20 but is different as this is importing a StructOpt from a different crate.

This issue relates to #307 

## Code

```rust
use structopt::StructOpt;

// This come from another crate
#[derive(Debug, StructOpt)]
enum BaseCli {
    Command1(Command1),
}

#[derive(Debug, StructOpt)]
struct Command1 {
    argument1: String,
}

// This is the structopt for this crate
#[derive(Debug, StructOpt)]
enum ExtendedCli {
    #[structopt(flatten)]
    BaseCli(BaseCli),
    Command2(Command2),
}

#[derive(Debug, StructOpt)]
struct Command2 {
    argument2: String,
}

fn main() {
    let cli = ExtendedCli::from_args();
    println!("{:?}", cli);
}
```

## Expected Result

```
test-struct-opt 0.1.0

USAGE:
    test-struct-opt <SUBCOMMAND>

FLAGS:

SUBCOMMANDS:
    command1
    command2
    help        Prints this message or the help of the given subcommand(s)
```

## Actual Result

```
error: flatten is only allowed on fields
  --> src/main.rs:15:17
   |
15 |     #[structopt(flatten)]
   |                 ^^^^^^^
```

cecton:
This is probably wrong but it works!

Fixes #327


cecton:
@CreepySkeleton OK with all :grin: I was kinda expecting it's no good.

Just one small question: in all your explanations, did you keep in mind that the structopt `BaseCli` and `ExtendedCli` (see #327) are in 2 different crates? It's very important.

CreepySkeleton:
@cecton It doesn't matter what crate they from as long as they both `derive(StructOpt)`. Dynamic dispatch doesn't care about it, AFAIK, feel free to correct me.

cecton:
@CreepySkeleton what I wish I could do is:

Instead of calling `augment_clap()` on a subcommand freshly created, I would call `augment_clap()` on `app` (given in argument) instead.

```rust
impl ::structopt::StructOptInternal for GroupCli {
    fn augment_clap<'a, 'b>(app: ::structopt::clap::App<'a, 'b>) -> ::structopt::clap::App<'a, 'b> {
        app.subcommand({
            let subcommand = ::structopt::clap::SubCommand::with_name("base-cli");
            let subcommand = {
                let kind = "FlattenStruct";
                let subcommand =
                    <BaseCli as ::structopt::StructOptInternal>::augment_clap(subcommand); // <------- HERE
                if <BaseCli as ::structopt::StructOptInternal>::is_subcommand() {
                    subcommand.setting(::structopt::clap::AppSettings::SubcommandRequiredElseHelp)
                } else {
                    subcommand
                }
            };
            subcommand.version("0.1.0")
        })
        .subcommand({
            let subcommand = ::structopt::clap::SubCommand::with_name("command3");
            let subcommand = {
                let kind = "other";
                let subcommand =
                    <Command3 as ::structopt::StructOptInternal>::augment_clap(subcommand);
                if <Command3 as ::structopt::StructOptInternal>::is_subcommand() {
                    subcommand.setting(::structopt::clap::AppSettings::SubcommandRequiredElseHelp)
                } else {
                    subcommand
                }
            };
            subcommand.version("0.1.0")
        })
        .version("0.1.0")
    }
```

But I don't think this is possible because `.subcommand()` moves `self` and return a new `App` with the new subcommands.

The only way I think it would be possible is to change this code:
```rust
    quote! {
        fn augment_clap<'a, 'b>(
            app: ::structopt::clap::App<'a, 'b>
        ) -> ::structopt::clap::App<'a, 'b> {
            app #app_methods #( #subcommands )* #version
        }
    }
```

And add at the end `#augments` that would be a list of augments to apply on `app`.

What do you think?

CreepySkeleton:
@cecton I think your current approach is what stalling you here. 

Current implementation generates a chain of `.subcommand().subcommand() //...` calls. What you're trying to do is putting the `augment_clap()` delegation somewhere inside this invocation, but `augment_clap` appends (potentially) multiple subcommands rather than only one. It can't live inside a `.subcommand()` call.

(I admit the previous `subcommands()` hack would work, but still no way we're accepting this 😋)

What you *want* to do is to generate *either* `.subcommand()` call (for normal subcommands) or `augment_clap()` call (for flattened subcommands) on every iteration. 

I recommend you to abandon the `.subcommand().subcommand()` approach here and generate this instead:
```rust
impl ::structopt::StructOptInternal for GroupCli {
    fn augment_clap<'a, 'b>(app: ::structopt::clap::App<'a, 'b>) -> ::structopt::clap::App<'a, 'b> {
        // normal subcommand
        let app = app.subcommand({
            normal subcommand
        });
 
        // flattened subcommand
        let app = <Sybty as StructOptInternal>::augment_clap(app);
     
        // and so on
    }

**Comment by cecton** [comment]
@CreepySkeleton I think I did it! It's probably still wrong at other levels and needs quite a clean-up but it works!

```rust
impl ::structopt::StructOptInternal for GroupCli {
    fn augment_clap<'a, 'b>(app: ::structopt::clap::App<'a, 'b>) -> ::structopt::clap::App<'a, 'b> {
        let app = app
            .subcommand({
                let subcommand = ::structopt::clap::SubCommand::with_name("command3");
                let subcommand = {
                    let kind = "other";
                    let subcommand =
                        <Command3 as ::structopt::StructOptInternal>::augment_clap(subcommand);
                    if <Command3 as ::structopt::StructOptInternal>::is_subcommand() {
                        subcommand
                            .setting(::structopt::clap::AppSettings::SubcommandRequiredElseHelp)
                    } else {
                        subcommand
                    }
                };
                subcommand.version("0.1.0")
            })
            .version("0.1.0");
        let app = <BaseCli as ::structopt::StructOptInternal>::augment_clap(app);
        app
    }
    fn from_subcommand<'a, 'b>(
        sub: (&'b str, Option<&'b ::structopt::clap::ArgMatches<'a>>),
    ) -> Option<Self> {
        if let Some(res) = <BaseCli as ::structopt::StructOptInternal>::from_subcommand(sub) {
            return Some(GroupCli::BaseCli(res));
        };
        match sub {
            ("command3", Some(matches)) => {
                let kind = "other";
                Some(GroupCli::Command3(
                    <Command3 as ::structopt::StructOpt>::from_clap(matches),
                ))
            }
            _ => None,
        }
    }
    fn is_subcommand() -> bool {
        true
    }
}
```

structopt-derive/src/lib.rs
```diff
@@ -458,14 +458,20 @@ fn gen_augment_clap_enum(
...
         let app_var = Ident::new("subcommand", Span::call_site());
         let arg_block = match variant.fields {
             Named(ref fields) => gen_augmentation(&fields.named, &app_var, &attrs),
             Unit => quote!( #app_var ),
             Unnamed(FieldsUnnamed { ref unnamed, .. }) if unnamed.len() == 1 => {
                 let ty = &unnamed[0];
+                let kind = match &*kind {
+                    Kind::FlattenStruct => "FlattenStruct",
+                    _ => "other",
+                };
```
CreepySkeleton: Is this some artifact you forgot to remove?

structopt-derive/src/lib.rs
```diff
@@ -449,15 +449,27 @@ fn gen_augment_clap_enum(
...
+            );
+            (variant, attrs)
+        })
+        .partition(|(_, attrs)| {
+            let kind = attrs.kind();
+            match &*kind {
+                Kind::FlattenStruct => true,
+                _ => false
+            }
+        });
```
CreepySkeleton: There's no need to separate flattened subcommands from the normal ones. I think it should look like this:
```rust
let subcommands = variants.iter().map(|variant| {
    // ...
    if let Kind::Flatten = kind {
        quote! {
            let app = augment_clap(app);
        }
    } else {
        quote! {
            let app = app.subcommand({
                // just as before
            });
        }
    }
});
```

CreepySkeleton: The same story with `from_subcommand`

cecton: Ok that one for `gen_augment_clap_enum` is fine but it doesn't work for `gen_from_subcommand` because they are match arms.

cecton: Ok I made it work without the match but... meh...
```rust
    fn from_subcommand<'a, 'b>(
        sub: (&'b str, Option<&'b ::structopt::clap::ArgMatches<'a>>),
    ) -> Option<Self> {
        if let Some(res) = <BaseCli as ::structopt::StructOptInternal>::from_subcommand(sub) {
            return Some(GroupCli::BaseCli(res));
        }
        if let ("command3", Some(matches)) = sub {
            return Some(GroupCli::Command3(
                <Command3 as ::structopt::StructOpt>::from_clap(matches),
            ));
        };
        None
    }
```

structopt-derive/src/lib.rs
```diff
@@ -543,15 +573,31 @@ fn gen_from_subcommand(
...
+        match variant.fields {
+            Unnamed(ref fields) if fields.unnamed.len() == 1 => {
+                let ty = &fields.unnamed[0];
+                quote! {
+                    if let Some(res) = <#ty as ::structopt::StructOptInternal>::from_subcommand(sub) {
+                        return Some(#name :: #variant_name (res));
+                    }
+                }
+            },
+            _ => todo!(),
```
cecton: @CreepySkeleton I left a few todo! also. I'm not sure what to do with those

cecton:
And what do I do with the failing test? Is there a way to distinguish a struct from an enum?

structopt-derive/src/lib.rs
```diff
@@ -458,39 +458,55 @@ fn gen_augment_clap_enum(
...
+        match &*kind {
+            Kind::Flatten => {
+                match variant.fields {
+                    Unnamed(FieldsUnnamed { ref unnamed, .. }) if unnamed.len() == 1 => {
+                        let ty = &unnamed[0];
+                        quote! {
+                            let app = <#ty as ::structopt::StructOptInternal>::augment_clap(app);
                         }
                     }
+                    _ => todo!(),
```
CreepySkeleton:
```suggestion
                    _ => abort!(variant.span(), "`flatten` is usable only with single-typed tuple variants"), 
```

structopt-derive/src/lib.rs
```diff
@@ -530,32 +548,49 @@ fn gen_from_subcommand(
...
+                match variant.fields {
+                    Unnamed(ref fields) if fields.unnamed.len() == 1 => {
+                        let ty = &fields.unnamed[0];
+                        quote! {
+                            if let Some(res) = <#ty as ::structopt::StructOptInternal>::from_subcommand(sub) {
+                                return Some(#name :: #variant_name (res));
+                            }
+                        }
+                    },
+                    _ => todo!(),
```
CreepySkeleton:
```suggestion
                    _ => abort!(variant.span(), "`flatten` is usable only with single-typed tuple variants"), 
```

structopt-derive/src/lib.rs
CreepySkeleton: Maybe I could have gone with this `if else` approach but I really don't like that flattened subcommands are prioritized over normal ones. Also... like you said, meh...

Let's back up for the partition and then something like 
```rust
match sub {
    // normal subcommands ...
    ("command3") => Some(GroupCli::Command3(
        <Command3 as ::structopt::StructOpt>::from_clap(matches),
    )),

    other => {
        // you can use if ... else if ... else if you want
        <FlattenedType as StructOpt>::from_subcommand(other)
            .or_else(|| <AnotherFlattenedType as StructOpt>::from_subcommand(other))
            .or_else(|| /* ... */)
            .map(|res| GroupCli::BaseCli(res))
    }
}
```

structopt-derive/src/lib.rs
cecton: done

structopt-derive/src/lib.rs
cecton: done

structopt-derive/src/lib.rs
```diff
@@ -500,7 +516,9 @@ fn gen_augment_clap_enum(
         fn augment_clap<'a, 'b>(
             app: ::structopt::clap::App<'a, 'b>
         ) -> ::structopt::clap::App<'a, 'b> {
-            app #app_methods #( #subcommands )* #version
+            let app = app #app_methods #version;
+            #( #subcommands )*;
+            app
```
CreepySkeleton: `#version` must go after `#subcommands` here due to another bug

tests/ui/struct_flatten.rs
```diff
@@ -9,10 +9,10 @@
...
-struct Opt {
-    #[structopt(short)]
-    s: String,
+#[structopt(name = "basic")]
+enum Opt {
+    #[structopt(flatten)]
+    Variant1,
 }
 
 fn main() {
```
CreepySkeleton: Simply delete this test along with the corresponding `.stderr`. There's no way to distinguish between a enum variant an struct.

structopt-derive/src/lib.rs
cecton: done

tests/ui/struct_flatten.rs
cecton: This is fine. I replaced it with the one with the error message you provided

structopt-derive/src/lib.rs
cecton: done

tests/ui/struct_flatten.rs
CreepySkeleton: Hmm, that is not what we testing here. The test is titled as `struct_flatten` so it's supposed to check `flatten ` on top of a struct. If you want to test an enum - rename the file.

structopt-derive/src/lib.rs
CreepySkeleton: There's the fancy grey `resolve Conversation` button next to such comments; you should hit it instead of wasting energy on typing "done", let's be energy-efficient!

tests/ui/struct_flatten.rs
cecton: Good point

cecton: done

CreepySkeleton: The button! It's really painful to see someone types useless messages...

structopt-derive/src/lib.rs
cecton: I thought it was for the reviewer to decide if this is resolved or not :grin:

cecton:
> I would like to highly encourage you to add a test so you can see if it works for youself.

Of course. I added one already. I think only the documentation is missing at this point.

structopt-derive/src/lib.rs
CreepySkeleton: Don't you worry, I have the even fancier `Unresolve conversation` button

CreepySkeleton:
Yes, and adjust the `.stderr` file

src/lib.rs
```diff
@@ -867,6 +867,40 @@
 //! }
 //! ```
 //!
+//! ### Flattening enums of subcommands
```
CreepySkeleton:
```suggestion
//! ### Flattening subcommands
```

src/lib.rs
```diff
@@ -867,6 +867,40 @@
...
+//! #[derive(StructOpt)]
+//! enum Opt {
+//!     #[structopt(flatten)]
+//!     BaseCli(BaseCli),
+//!     Command2(Command2),
+//! }
+//! ```
+//!
+//! + `cli command1 42`
+//! + `cli command2 42`
```
CreepySkeleton: Please use something like
\```shell
cli command1 42
cli command2 42
\```

And also, could you please try to come up with something more fancy than `command1,2`? They are too similar :(

structopt-derive/src/lib.rs
```diff
@@ -458,39 +458,55 @@ fn gen_augment_clap_enum(
...
+                                    #app_var.setting(
+                                        ::structopt::clap::AppSettings::SubcommandRequiredElseHelp
+                                    )
+                                } else {
+                                    #app_var
+                                }
+                            }
+                        }
+                    }
+                    Unnamed(..) => abort_call_site!("{}: tuple enums are not supported", variant.ident),
```
CreepySkeleton: Please avoid `abort_call_site!` whenever you have a span to point to.

```suggestion
                    Unnamed(..) => abort!(variant.span(), "non single-typed tuple enums are not supported"),
```

cecton: That git revert didn't ended how I expected :grin:

CreepySkeleton:
You need `cargo fmt --all`

src/lib.rs
CreepySkeleton: So, what about `command1,2` stuff? It's really sad that I'm the only one who comes up with interesting names 😭

CreepySkeleton:
hey @TeXitoi mind to take a look?

src/lib.rs
```diff
@@ -867,6 +868,42 @@
...
+//! # fn main() {}
+//! #[derive(StructOpt)]
+//! enum BaseCli {
+//!     Ghost10(Ghost10),
+//! }
+//!
+//! #[derive(StructOpt)]
+//! struct Ghost10 {
+//!     arg1: i32,
+//! }
```
TeXitoi: You can use "enum struct". It's quite convenient in examples as it is shorter.

```suggestion
//! enum BaseCli {
//!     Ghost10 { arg1: i32 },
//! }
```

tests/ui/enum_flatten.rs
```diff
@@ -9,10 +9,10 @@
 use structopt::StructOpt;
 
 #[derive(StructOpt, Debug)]
-#[structopt(name = "basic", flatten)]
```
TeXitoi: This test should not be removed. I think you might add a new test instead of modifying this one.

CreepySkeleton:
Fuck, I should have waited for this to get merged before merging my own PR. 

OK, once @TeXitoi has it reviewed, I'll deal with the conflicts myself as part of my apologies to @cecton , sorry 😢

tests/ui/enum_flatten.rs
CreepySkeleton: Unfortunately, it's really hard to distinguish between an attr on top of enum variant (valid usage) and on top of struct/enum (invalid usage) in current codebase. I think this test is not so important and the changes required to do it correctly would be quite complex so I propose to ditch this test for now.

TeXitoi: So this code now compiles? What does it do?

CreepySkeleton: Flatten is simply ignored

TeXitoi: OK, then

PR #337, opened 2020-01-21
Suppress all clippy lints except correctness

CreepySkeleton:
The correctness category is for code that's outright wrong. I think we should not generate such code and the extra check won't hurt.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,7 @@ matrix:
 
 script:
   - cargo test -- -D warnings
+
+jobs:
+  allow_failures:
+    - name: clippy
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
     - rust: nightly
 
 script:
-  - cargo test -- -D warnings
+  - cargo test
 
 jobs:
   allow_failures:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -17,10 +17,19 @@ matrix:
...
     - rust: beta
     - rust: nightly
 
 script:
-  - cargo test $FEATURES
+  - cargo test
+
+jobs:
+  allow_failures:
+    - name: clippy
```
TeXitoi: Why allowing failures?

src/lib.rs
```diff
@@ -349,7 +349,6 @@
 //! ```
 //! # extern crate clap;
 //! # fn parse_validator<T>(_: String) -> Result<(), String> { unimplemented!() }
-//! # fn main() {
```
TeXitoi: The extern crate line should not be needed.

src/lib.rs
```diff
@@ -426,10 +424,8 @@
 //!
 //! }
 //!
-//! # fn main() {
 //! # Opt::from_clap(&Opt::clap().get_matches_from(
```
TeXitoi: `Opt::from_iter` may be better here.

.travis.yml
CreepySkeleton: Because clippy checks not only structopt but also all of its dependencies. BlockingCI just because of some lint in one of the deps would be senseless.

PR #340, opened 2020-01-23
Support for hidden subcommands

TyPR124:
Adds support for a `#[structopt(hidden)]` attribute which hides subcommands (they don't show up in help) when present on an enum variant.

Currently it is possible to use this attribute anywhere, but it shouldn't do anything except when attached to an enum variant.

Was hoping to get some feedback and if this is something that can be merged I can fix up anything that needs fixing.

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -89,6 +89,7 @@ pub struct Attrs {
     verbatim_doc_comment: Option<Ident>,
     has_custom_parser: bool,
     kind: Sp<Kind>,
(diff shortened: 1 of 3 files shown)


CreepySkeleton:
You can use it via raw method:

```rust
#[derive(StructOpt)]
enum Sub {
    Add,
    #[structopt(setting = AppSettings::Hidden)]
    Remove
}
```

There's no need for `#[structopt(hidden)]`.

CreepySkeleton:
But thanks anyway!

PR #342, opened 2020-01-24
Fix verbatim_doc_comment on fields

davidMcneil:
Signed-off-by: David McNeil <mcneil.david2@gmail.com>


CreepySkeleton:
Good catch, thanks!

PR #353, opened 2020-02-29
`clone-impls` missing from syn usage in structopt_derive

palfrey:
While messing around with trying to fix something else, I managed to run into a series of issues in `structopt_derive` of the form

```rust
error[E0277]: the trait bound `syn::expr::Expr: std::clone::Clone` is not satisfied
  --> src/attrs.rs:28:10
   |
28 |     Skip(Option<Expr>),
   |          ^^^^^^^^^^^^
   |          |
   |          expected an implementor of trait `std::clone::Clone`
   |          help: consider borrowing here: `&Option<Expr>`
   |
   = note: required because of the requirements on the impl of `std::clone::Clone` for `std::option::Option<syn::expr::Expr>`
   = note: required by `std::clone::Clone::clone
```

This PR adds a explicit build step of `structopt_derive` on it's own to demostrate the bug (see also https://travis-ci.org/TeXitoi/structopt/builds/656783840) and then adds the `clone-impls` feature to the `syn` features in `structopt-derive` to fix this. I think the build was working before because something else in `structopt` included syn and included the relevant feature, but `structopt_derive` didn't explicitly add it itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
     - rust: nightly
 
 script:
-  - cargo build
+  - cd structopt-derive && cargo build && cd ..
   - cargo test
 
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ matrix:
     - rust: nightly
 
 script:
+  - cargo build
   - cargo test
 
 jobs:
--- a/structopt-derive/Cargo.toml
+++ b/structopt-derive/Cargo.toml
@@ -14,7 +14,7 @@ license = "Apache-2.0/MIT"
 travis-ci = { repository = "TeXitoi/structopt" }
 
 [dependencies]
-syn = { version = "1", default-features = false, features = ["derive", "parsing", "proc-macro"] }
+syn = { version = "1", default-features = false, features = ["derive", "parsing", "proc-macro", "clone-impls"] }
 quote = "1"
 proc-macro2 = "1"
 heck = "0.3.0"


BeatButton:
+1, just spend a couple of hours trying to figure this out

.travis.yml
```diff
@@ -28,6 +28,7 @@ matrix:
     - rust: nightly
 
 script:
+  - cd structopt-derive && cargo build && cd ..
```
CreepySkeleton: Why do you think this is needed?

CreepySkeleton:
> I think the build was working before because something else in structopt included syn and included the relevant feature, but structopt_derive didn't explicitly add it itself.

Yeah, I released a new version of `proc-macro-error` several hours ago, and now it doesn't require `clone-impls`. Oops, we overlooked it.

CreepySkeleton:
Btw, I*did* check the release with structopt locally, and it *passed*. WTF???

CreepySkeleton:
Fixed by #355 and published as `0.3.10`

CreepySkeleton:
> Btw, I did check the release with structopt locally, and it passed. WTF???

Hm, I guess `cargo clean` so cargo simply used the old cached artifact since I'd forgotten `cargo clean`. Lesson learned.

.travis.yml
palfrey: To both reproduce the problem and identify anything like it in the future

CreepySkeleton: cargo build builds both crates anyway, so it's kind of meaningless 🤷‍♂

palfrey: Except that if structopt includes a feature that structopt-derive needs (as happened here) we won't be able to spot the problem. This build step means we can be sure structopt-derive is capable of compiling correctly on it's own.

CreepySkeleton: >  structopt includes a feature that structopt-derive needs

That's not what happened here. `structopt-derive` (not `sturctopt`!) depends on `proc-macro-error`, which used to depend on `rustversion` up until the yesterday's release, which enables the "full" feature. I've ditched `rustversion` yesterday, so the dependency on `proc-macro-error` doe not result in `syn-full` enabled anymore. As you can see, if `structopt-derive` fails to build, structopt would fail too.

`structopt` on it's own is innocent here, this was caused by the derive and derive only.

CreepySkeleton: Disregard my last, I was talking about `cargo test`, not `cargo build`. But still, I doubt this problem will ever arise again (yeah, sure).

PR #383, opened 2020-04-25
Allow everything that implements ToString as default_value and env

CreepySkeleton:
Implemented handling of `default_value` and `env`. Maybe we could do the other methods too, but we were ever asked only about these two.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+# [Upcoming]
+
+* Now everything that implements `ToString` can be passed to `#[structopt(default_value/env)]`.
+
+  ```rust
+  #[derive(StructOpt, PartialEq, Debug)]
+  struct Opt {
+      #[structopt(default_value = 2 + 2 * 2)]
+      arg: i32,
+  }
+  assert_eq!(Opt { arg: 6 }, Opt::from_iter(&["test"]));
+  assert_eq!(Opt { arg: 1 }, Opt::from_iter(&["test", "1"]));
+  ```
+
 # v0.3.14 (2020-04-22)
 
 * Minor documentation improvements.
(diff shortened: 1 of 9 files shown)


TeXitoi:
Mmm, missunderstanding...

I mean:

```rust
use structopt::StructOpt;

macro_rules! into_str {
    ($x:expr) => {{
        structopt::lazy_static::lazy_static! {
            static ref DATA: String = $x.to_string();
        }
        DATA.as_str()
    }};
}

#[derive(Debug, StructOpt)]
struct Opt {
    #[structopt(default_value = into_str!(2+5))]
    number: i32,
}
fn main() {
    let cmd = Opt::from_args();
    println!("{:#?}", cmd);
}
```

So no needed modification, usable for all the args, just a helper macro provided by the structopt crate.

CreepySkeleton:
Meh, why not? Already done, doesn't break anything, super easy to use, pretty easy to implement.

TeXitoi:
I can see 2 downside:
 - if it works for `default_value`, the user will not understand why it doesn't work for the other methods that takes `&str`.
 - today, the structopt users can give something to `default_value` that is not a string literal. I can think at least of a static `&str`, and a lazy_static. These examples might not break, but they will allocate one useless time. And maybe it can break in some case, didn't find an example, but I will not be surprised if someone find one.

CreepySkeleton:
Well, it cannot break because everything that people could possibly use here must have been `&str`, because those methods explicitely take `&str`. It has to be `&str` (or another reference that [`impls Deref<str>`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=34b3f704494b9a54dd6b9d2d29ee67724)) in order to compile in the first place. 

We use `.to_string().as_str()` which always return `&str`. Therefore, if it implements `ToString`, it won't break. And `&str` (or something that derefs to it) does implement `ToString`. 

But yeah, the other points stand. I don't really feel like implementing the hack for every single method and I don't like the double allocation either.

So what do we do?

CreepySkeleton:
@TeXitoi do we close it? I'd say yes

TeXitoi:
OK, let's close this. Do you want I do a PR with the macro?

Issue #405, opened 2020-07-08
Add external-doc feature


swfsql:
Rust has an unstable feature [external-doc](https://doc.rust-lang.org/beta/unstable-book/language-features/external-doc.html) which could have an effect in structopt, such as for:

```rust
#[doc(include = "../README.md")]
struct Args {}
```

I think this could be implemented by declaring usage of such unstable feature by a library-gated feature, and then change the appropriate doc implementation.

The current behaviour also could stay intact by diverging the implementation accordingly to the library-gated feature.

swfsql:
Given the reasoning and conclusion at https://github.com/TeXitoi/structopt/pull/406#pullrequestreview-446230548 - external-doc is not advisable for usage as help documents.

PR #406, opened 2020-07-08
init external_doc feature

swfsql:
Initial implementation for https://github.com/TeXitoi/structopt/issues/405

Adds a library-gated feature, which forwards it to the derive library, which declares usage of the unstable feature and then implements it on `structopt-derive/src/attrs.rs`.

This currently doesn't adds anything to the tests nor documentation..  
In the short run, personally I plan on using this PR as a patch for structopt.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ debug = ["clap/debug"]
 no_cargo = ["clap/no_cargo"]
 doc = ["clap/doc"]
 paw = ["structopt-derive/paw"]
+external_doc = ["structopt-derive/external_doc"]
 
 [badges]
 travis-ci = { repository = "TeXitoi/structopt" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ debug = ["clap/debug"]
 no_cargo = ["clap/no_cargo"]
 doc = ["clap/doc"]
 paw = ["structopt-derive/paw"]
-external_doc = ["structopt-derive/external_doc"]
+unstable_external_doc = ["structopt-derive/unstable_external_doc"]
 
 [badges]
 travis-ci = { repository = "TeXitoi/structopt" }
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -371,14 +371,49 @@ impl Attrs {
         let comment_parts: Vec<_> = attrs
             .iter()
             .filter(|attr| attr.path.is_ident("doc"))
-            .filter_map(|attr| {
-                if let Ok(NameValue(MetaNameValue { lit: Str(s), .. })) = attr.parse_meta() {
-                    Some(s.value())
-                } else {
-                    // non #[doc = "..."] attributes are not our concern
-                    // we leave them for rustc to handle
-                    None
+            .filter_map(|attr| match attr.parse_meta() {
+                Ok(NameValue(MetaNameValue { lit: Str(s), .. })) => Some(s.value()),
+                Ok(NameValue(_)) => None,
+                Ok(Path(_)) => None,
+                // external-doc
+                // eg. #[doc(include = "../README.md")]
+                #[cfg(feature = "external_doc")]
+                Ok(List(syn::MetaList { nested, .. })) => {
+                    let content: String = nested
+                        .iter()
+                        .filter_map(|nested_meta| match nested_meta {
+                            syn::NestedMeta::Meta(syn::Meta::NameValue(nv)) => {
+                                if nv.path.is_ident("include") {
+                                    match &nv.lit {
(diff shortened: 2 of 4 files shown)


structopt-derive/src/lib.rs
```diff
@@ -11,6 +11,7 @@
 //! for the usage of `#[derive(StructOpt)]`.
 
 #![allow(clippy::large_enum_variant)]
+#![cfg_attr(feature = "external_doc", feature(external_doc))]
```
CreepySkeleton, line 14: Why would we need that?

Cargo.toml
```diff
@@ -22,6 +22,7 @@ debug = ["clap/debug"]
 no_cargo = ["clap/no_cargo"]
 doc = ["clap/doc"]
 paw = ["structopt-derive/paw"]
+external_doc = ["structopt-derive/external_doc"]
```
CreepySkeleton, line 25: unstable features might be, well, unstable :) I would like the fact to be reflected in the name of the feature, like `unstable_external_doc` or something so users are aware this is subject to change/break.

structopt-derive/src/attrs.rs
```diff
@@ -371,14 +371,49 @@ impl Attrs {
...
+                        .filter_map(|nested_meta| match nested_meta {
+                            syn::NestedMeta::Meta(syn::Meta::NameValue(nv)) => {
+                                if nv.path.is_ident("include") {
+                                    match &nv.lit {
+                                        Str(s) => {
+                                            let path = std::path::PathBuf::from("src")
+                                                .join(s.value())
+                                                .canonicalize()
+                                                .unwrap();
+                                            Some(std::fs::read_to_string(&path).unwrap())
```
CreepySkeleton, line 393: Please use `abort!` instead of `unwrap`s here

structopt-derive/src/lib.rs
swfsql: This is needed so the unstable feature `external_doc` is not unnecessarily required, as it would by itself require nightly rustc. So instead of doing `#![feature(external_doc)]`, I used that other attr instead.

structopt-derive/src/attrs.rs
swfsql: Changed it! Sorry for not checking the library culture; and thanks for indicating it (it is very cool)

Cargo.toml
swfsql: I agree, and made the change to reflect that

swfsql:
> What is you motivation for that, anyway? `App::help("...")` is supposed to be _reasonably concise_ description for the application, and extracting it to an external file isn't exactly the best way to achieve this.

I had a couple CLI that had quite a few changes quite often, so instead of changing both the Readme and the help documentation, I preferred to had a single place of change.  
But personally I didn't include "readme.md" file; I included instead a particular document (usage.md), and referenced that document both from README.md and from structopt help documentation.

But anyway I did still insert a short introduction onto the README (that would not appear on the CLI help usage), and also insert a short introduction as doc-comments (`///`) onto the Args struct (that would not appear on the README) as well - I just needed to add the doc-comments before the `#[doc(include = "..")]`.

---

Having said that, actually I thought that this form of `#[doc]` were also included even before trying it.. at first I just naturally had this expectation..

---

But after having a (somehow middle-sized) "about (markdown) document", I got to wonder how the --help would look if bat syntax highlighting were used as a printer.. !

structopt-derive/src/lib.rs
CreepySkeleton: I wasn't asking why it's behind `cfg_attr`, I was asking why _structopt_ needs `#[feature(external_doc)]` in the first place :) We don't use `#[doc(include = "...")]` anywhere so we shouldn't need it at all

swfsql: You're right, that unstable feature is always unnecessary. Thanks!

swfsql:
> That's a lot of headache, and I've got long backlog already, so the solution is entirely in your hands.

> I am very hesitant supporting something that I might need to come back and redo, especially considering that structopt is on its way to be soft-deprecated.

This is true. In the end actual markdown files don't work well. And I wouldn't be maintaining anything.

> Last but not least: are you sure that #[doc(include)] expands after derive macros do? Is this specified? Did you test the patch?

I expect it to resolve in the order that they appear on the code.

This:
```rust
/// a  
/// b  
/// c
#[doc(include = "../about.md")] // d f g
struct Args {}
```

outputs:
```
structopt-external-doc 0.1.0
a b c

d e f

USAGE: ...
```

You can test it with:
```bash
git clone https://github.com/swfsql/structopt-external-doc.git
cd structopt-external-doc
cargo run -- --help
```

---

Having said that, I believe that this can be closed.

PR #408, opened 2020-07-10
Fix typo in required_if usage in crate docs

rgardner:
The `required_if` example in `lib.rs` doesn't actually require `file_name` when `out_type=file` because the arg name is wrong: clap uses hyphens for word separators, not underscores. This change fixes the example to use `out-type`. I decided against adding an inline test to keep the crate docs succinct.

I also added a separate `required_if` example with tests to validate this behavior. I separated this change into a separate commit in case you just want the typo fix without the new example.

Thank you for structopt!

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -69,6 +69,10 @@ How to completely remove version.
 
 How `#[structopt(rename_all)]` works.
 
+### [Required If](required_if.rs)
+
+How to use `#[structopt(required_if)]`.
+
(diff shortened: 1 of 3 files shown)


CreepySkeleton:
That's great, thanks! 

I think we'll take the example, it's a good one :)

rgardner:
Awesome, thanks @CreepySkeleton.

PR #412, opened 2020-07-24
Add support for lower/upper in rename_all

Morganamilo:
Some programs do not use anything to separate word boundaries.

For example a struct may contain the field `build_dir` while the flag is
`--builddir`.

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -65,6 +65,10 @@ pub enum CasingStyle {
     Snake,
     /// Use the original attribute name defined in the code.
     Verbatim,
+    /// Keep all letters lowercase and remove word boundaries.
+    Lower,
+    /// Keep all letters uppercase and remove word boundaries.
+    Upper,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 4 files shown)


CreepySkeleton:
@ping @TeXitoi

structopt-derive/src/attrs.rs
```diff
@@ -208,6 +214,8 @@ impl Name {
                     ScreamingSnake => s.to_shouty_snake_case(),
                     Snake => s.to_snake_case(),
                     Verbatim => s,
+                    Lower => s.to_lowercase().replace("_", ""),
+                    Upper => s.to_uppercase().replace("_", ""),
```
TeXitoi, line 218:
```suggestion
                    Lower => s.to_snake_case().replace("_", ""),
                    Upper => s.to_shouty_snake_case().replace("_", ""),
```

We can't rely on snake case for the origin.

Morganamilo: What other case could there by? Camel case would be converted when lowering and `-` are not allowed in variable names.

Morganamilo: Thinking about it, i forgot this also applied so environment variables so '-' could indeed be present.

CreepySkeleton: That reminded me - you have only changed `rename_all` but left `rename_all_env` untouched. I don't think this kind inconsistency does us any good; could you please tweak it as well?

CreepySkeleton: Now wait, it's changed. Nevermind.

PR #414, opened 2020-07-25
Allow #[flatten] fields to have doc comments but ignore them (comments)

### LINKED ISSUE #411: Documenting flattened fields?
I'm trying to write Cargo documentation for a field that is flattened by structopt. This seems to be rejected as a compile-time error, via:

```
  --> foo/src/serve.rs:12:10
   |
12 | #[derive(StructOpt, Debug)]
   |          ^^^^^^^^^
   |
   = help: message: methods and doc comments are not allowed for flattened entry

```

I imagine StructOpt chose this behavior for automatically generated `--help`, however this directly conflicts with clippy lints that require documentation. Ie, i can't merge this structopt code as it either fails clippy lints or fails to compile due to StructOpt.

Is there a way to disable this compilation failure? Ie, ignore the flattened field's documentation? Maybe something like a `structopt(flatten, ignore_documentation)` attribute?

CreepySkeleton:
Closes #411


tests/ui/flatten_and_methods.stderr
```diff
@@ -1,4 +1,4 @@
-error: methods and doc comments are not allowed for flattened entry
+error: methods are not allowed for flattened entry
```
TeXitoi, line 1: Wasn't the goal to ignore the doc  comment? Here, the compilation fail.

CreepySkeleton: The goal is to ignore the comments, but disallow methods with flatten, like `short`. This is exactly what triggers it here. 
```
/// Comments allowed
#[structopt(flatten)]

// Methods disallowed
#[structopt(flatten, short)]
```

PR #427, opened 2020-09-08
Add support for renaming argument values

chrisduerr:
I just found out that the number of values one can rename with `name =
"XXX"` is limited to one, since it's the unique argument value name.

Looking at the code and documentation, I could not find any way to set
clap's `value_name`, so this commit adds a new `value_name` attribute
which allows setting the name for a specific attribute.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,7 @@
 //!     - [External subcommands](#external-subcommands)
 //!     - [Flattening subcommands](#flattening-subcommands)
 //! - [Flattening](#flattening)
+//! - [Changing value names](#changing-value-names)
 //! - [Custom string parsers](#custom-string-parsers)
 //!
 //!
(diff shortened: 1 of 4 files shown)


chrisduerr:
The failure isn't related to this PR, so CI should be good.

CreepySkeleton:
It kinda does work as is, with the exact syntax your PR reintroduces. You can call any method from `clap::App/Arg`; it's called "raw methods" in documentation.

chrisduerr:
> it's called "raw methods" in documentation.

I can't find that anywhere in the structopt documentation. It does work though.

CreepySkeleton:
https://docs.rs/structopt/0.3.17/structopt/index.html#raw-methods

chrisduerr:
I must be going crazy because I could not find that for the life of me. Thanks for pointing it out.

PR #431, opened 2020-09-17
Friendlier example text in tutorial

toothbrush:
As a new Rust user, i did not find the snark against Git to be helpful.  It's probably a friendly joke, but it doesn't (IMHO) translate well in text.

My suggestion is to simply change this to a more neutral text to not alienate less experienced Rustaceans like myself :smiley: 

Thank you for your great library :raised_hands:

### DIFF
--- a/examples/git.rs
+++ b/examples/git.rs
@@ -7,7 +7,7 @@ use structopt::StructOpt;
 
 #[derive(StructOpt, Debug)]
 #[structopt(name = "git")]
-/// the stupid content tracker
+/// the well-known version control tool
 enum Opt {
     /// fetch branches from remote repository
     Fetch {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -764,7 +764,7 @@
 //!
 //! # use std::path::PathBuf;
 //! #[derive(StructOpt)]
-//! #[structopt(about = "the stupid content tracker")]
+//! #[structopt(about = "the well-known version control tool")]
 //! enum Git {
 //!     Add {
 //!         #[structopt(short)]


CreepySkeleton:
These words are how git presents itself in [its official main page](https://git-scm.com/docs/git). These exact words. Linus himself chose this description years ago, and it's been carefully and lovely carried on to this day. This isn't a joke (not ours, anyway), this is the official motto, as official as you get.

Whatever the case, I would agree with you if there was an insult in there, but this doesn't qualify as such for any widely accepted definition of insult. The word "stupid" is not an offense on it's own, or you would have to count offensive such phrases as "keep it simple, stupid", "the more stupid the better", "against stupidity God Himself is helpless" and so on.

I'll leave the final decision up to @TeXitoi , but I perceive this PR as an example of mindless wordhunting, completely ignoring context. If you want to make a real difference, you should go talk to Linus and ask him to change the motto.

TeXitoi:
I agree with @CreepySkeleton that, as it is the exact official documentation of git, it should stay as is.

Anyway, thanks for your interest in improving structopt.

toothbrush:
Ah, well today i learned something! I had no idea about this being the "official" way Git is described, so my apologies for displaying my ignorance of history 😊

TeXitoi:
No problem, the intention was good.

Waiting for your next contribution ;-)

toothbrush:
Having slept on it to avoid posting a rash response, i will leave some feedback for you to consider @CreepySkeleton.

You did not simply close this issue without replying, in fact i very much appreciated your explanation of where this snippet comes from.  It taught me something, as somebody who does not know this history, thanks for that!  However, i would gently suggest that you could have dropped the final paragraph:

> I'll leave the final decision up to @TeXitoi , but I perceive this PR as an example of mindless wordhunting, completely ignoring context. If you want to make a real difference, you should go talk to Linus and ask him to change the motto.

I feel like it is quite an uncharitable assessment of my intentions, and i think we can all strive towards a friendly and inclusive community.  If you leave off that final paragraph, the meaning and message is completely retained, but it saves me some unnecessary cortisol :smile_cat: 

Anyway, that's just my 2c, i'll leave it at that.  Thank you for maintaining a very useful library! :smiley:

CreepySkeleton:
It wasn't your intentions I was talking about. I was ninety nine percent sure your intentions were the best possible, I still am, and the fact that I went out of my way to give you a specific link to the man page should confirm my honesty. Your approach and lack of consideration were the problems.

You know, there was a somewhat lengthy comment here in which I tried to reflect on my thoughts and hopefully explain why I was being terse. Then I looked at it and removed everything except the intro: this kind of comments is almost always left unread in my experience. If you are interested in that, feel free to send me an email, and we will have a nice friendly conversation that will hopefully get us understand each other better. Or we can leave it at this.

Please don't reply here. This isn't the place, people get notified uselessly.

PR #434, opened 2020-09-23
Forbid unsafe

### LINKED ISSUE #432: Forbid unsafe code
`structopt` does not seem to use unsafe code, which is very good :)

To prevent this from ever happening, add following line to the top:
```rust
#![forbid(unsafe_code)]
```
of https://github.com/TeXitoi/structopt/blob/master/src/lib.rs
and https://github.com/TeXitoi/structopt/blob/master/structopt-derive/src/lib.rs

This will make the package green in 
https://github.com/rust-secure-code/cargo-geiger

CreepySkeleton:
Closes #432 

@TeXitoi Do you mind if I do a new patch release with this change? Just so `cargo geiger` would list structopt as green.


TeXitoi (approved):

TeXitoi:
OK for a new release, but you lack a changelog entry

CreepySkeleton:
Sure, will do in a "release" PR. I didn't include it here because you could say no 😄

PR #438, opened 2020-10-08
V0.3.19

CreepySkeleton:
@TeXitoi I accidentally pushed [two](https://github.com/TeXitoi/structopt/commit/f5a7dcc6fa2a98db59046fb1c93bea1f3880aba7) [commits](https://github.com/TeXitoi/structopt/commit/4486e5876c4f295a1e794e1e54d20c6db68d17a1) directly to master, sorry. Comments are good, no junk in the commit history, but maybe we should protect master so it can't be pushed to?


TeXitoi:
I've done something for the master branch, but not sure it disallow pushing.

TeXitoi (approved):

CreepySkeleton:
[Nope](https://github.com/TeXitoi/structopt/commit/98d5d80527e25999d7015239a6e911e0c02e8a5f) (and sorry for the noise, but this is the only way to know for sure). [This](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-branch-restrictions) should do the trick

TeXitoi:
That's what I've done, but I don't have "Restrict who can push to matching branches."

PR #453, opened 2020-12-12
Show "debug" in version for unoptimized build

prateeknischal:
For an unoptimized build, i.e. no `--release` flag, the binary does not
have a user friendly way to know if it's a debug build once transported
from the target folder.

This PR adds a "-debug" string when the binary is built with
"debug_assertions".

### DIFF
--- a/structopt-derive/src/attrs.rs
+++ b/structopt-derive/src/attrs.rs
@@ -602,9 +602,16 @@ impl Attrs {
             (None, Some(m)) => m.to_token_stream(),
 
             (None, None) => std::env::var("CARGO_PKG_VERSION")
-                .map(|version| quote!( .version(#version) ))
+                .map(|version| {
+                    let mut if_debug = version;
+                    if cfg!(debug_assertions) {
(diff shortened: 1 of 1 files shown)


TeXitoi:
We can't change the version like that, it would change the output of programs. And I don't think that's the good way of doing this, version is for, you know, version, not the kind of build.

prateeknischal:
I understand this would be more of a disruptive change since it's changing the behaviour of a core function. But versions do have characteristic information about the build right. I mean if you look at semantic versioning, it has a more descriptive tokens, major, minor, patch, build numbers, release identifiers etc. 

I thought its might be a good idea to include this info by default since i has used a debug version of the build and it was running kind of slow for rust. I later realized, I never built a release version for it. If I had this feature enabled, I would not have to worry about tagging my builds with debug or release versions but automatically get that info. 

I guess keeping it simple makes sense and if someone needed detailed info in their versions, there are crates for that. I shall close the PR in that case.

TeXitoi:
You can do that manually in your crate if you need, but we can't change that here, because some applications will change its behavior.

PR #479, opened 2021-05-18
tests pass with Rust v1.35

tshepang:


TeXitoi:
We don't want to support a older rust version: if we need to use a feature that is only in 1.36, we don't want to be blocked.

1.36 was the current version when v0.2 goes out.

1.35 might work, but we don't guaranty it, and doing this change will force us to guaranty it.

Thanks for your interest!

TeXitoi:
See https://github.com/TeXitoi/structopt#structopt-rustc-version-policy

TeXitoi:
That's also the (current) clap minimum version https://github.com/clap-rs/clap/blob/v2-master/.travis.yml#L29

Issue #128, opened 2018-07-26
Structopt derive doesn't like type-parametric structs


vorner:
Probably better showed with some code:

```rust
struct Wrapper<O> {
    #[structopt(flatten)]
    sub: O,
}
```

I get this error message:

```
error[E0412]: cannot find type `O` in this scope
  --> src/lib.rs:31:10
   |
31 | #[derive(StructOpt)]
   |          ^^^^^^^^^ did you mean `OW`?

error[E0243]: wrong number of type arguments: expected 1, found 0
  --> src/lib.rs:31:10
   |
31 | #[derive(StructOpt)]
   |          ^^^^^^^^^ expected 1 type argument
```

Similar thing works with serde, where it automatically places correct type bounds on the generated impl block.

TeXitoi:
StructOpt was not design with generics in mind. That's a complicated thing.

What is your use case?

vorner:
Yes, I already noticed it's a bit harder than just making sure the impl gets the type bounds, because some things are not part of the trait. But I've found a workaround for my own use case.

I'm building a helper start-me-a-daemon library. Some of the options will be specific for each daemon, while some others will be common for all of them (like logging and option not to go to background). So I want to compose the common ones (defined in the library) with the daemon-specific ones provided by the user of the library as a generic.

For now, I have this (I haven't tried it yet, but I guess it should work).

vorner:
Yes, but that's a different use case ‒ they want to expose the verbosity flag to the end application. I don't want to give the application the logging configuration etc ‒ I just want to logging to be configured automagically. That means my library needs access to the common configuration (if I went the verbosity way, I'd have to let the user tell me where the logging configuration is) and it would pollute the user's configuration structure.

vorner:
(Oh, I haven't included the link in the post up there ↑ ‒ this is what I do now: https://github.com/vorner/spirit/blob/master/src/lib.rs#L59)

TeXitoi:
Great workaround!

CreepySkeleton:
Let's get back to the original issue. Do we want to support generic structs and what the pitfalls are?
Or will we just close it as "wontfix"?

TeXitoi:
wontfix is fine for me.

CreepySkeleton:
[...]

PR #483, opened 2021-06-24
add basic support for generic structs and enums

njeffords:
This PR provides (at least basic) support for generics when deriving `StructOpt` (issue #128). It's probably not the ideal final implementation, but from what I can see, it improves the situation and makes some use cases possible.

### DIFF
--- a/structopt-derive/src/lib.rs
+++ b/structopt-derive/src/lib.rs
@@ -770,32 +770,54 @@ fn split_structopt_generics_for_impl(generics: &Generics) -> (ImplGenerics, Type
         return false;
     }
 
-    struct TraitBoundAmendments(TokenStream);
-    impl TraitBoundAmendments {
-        fn new() -> Self { Self(TokenStream::new()) }
+    struct TraitBoundAmendments{
+        tokens: TokenStream,
+        need_where: bool,
+        need_comma: bool,
+    }
 
-        fn extend(&mut self, amendment: TokenStream) {
-            if !self.0.is_empty() {
-                self.0.extend(quote!{ , });
+    impl TraitBoundAmendments {
+        fn new(where_clause: Option<&WhereClause>) -> Self {
+            let tokens = TokenStream::new();
+            let (need_where,need_comma) = if let Some(where_clause) = where_clause {
+                if where_clause.predicates.trailing_punct() {
+                    (false, false)
+                } else {
+                    (false, true)
+                }
+            } else {
+                (true, false)
+            };
+            Self{tokens, need_where, need_comma}
+        }
+
+        fn add(&mut self, amendment: TokenStream) {
+            if self.need_where {
+                self.tokens.extend(quote!{ where });
+                self.need_where = false;
             }
-            self.0.extend(amendment);
+            if self.need_comma {
+                self.tokens.extend(quote!{ , });
+            }
+            self.tokens.extend(amendment);
+            self.need_comma = true;
         }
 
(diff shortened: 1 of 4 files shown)


structopt-derive/src/lib.rs
```diff
@@ -748,19 +748,83 @@ fn gen_paw_impl(name: &Ident) -> TokenStream {
...
 #[cfg(not(feature = "paw"))]
-fn gen_paw_impl(_: &Ident) -> TokenStream {
+fn gen_paw_impl(_: &ImplGenerics, _: &Ident, _: &TypeGenerics, _: &TokenStream) -> TokenStream {
     TokenStream::new()
 }
 
+fn split_structopt_generics_for_impl(generics: &Generics) -> (ImplGenerics, TypeGenerics, TokenStream) {
+    use syn::{ token::Add, TypeParamBound::Trait };
+
+    fn path_is_structop(path: &Path) -> bool {
```
TeXitoi, line 758:
```suggestion
    fn path_is_structopt(path: &Path) -> bool {
```

structopt-derive/src/lib.rs
```diff
@@ -748,19 +748,83 @@ fn gen_paw_impl(name: &Ident) -> TokenStream {
...
+        path.segments.last().unwrap().ident == "StructOpt"
+    }
+
+    fn type_param_bounds_contains_structop(bounds: &Punctuated<TypeParamBound, Add>) -> bool {
```
TeXitoi, line 762:
```suggestion
    fn type_param_bounds_contains_structopt(bounds: &Punctuated<TypeParamBound, Add>) -> bool {
```

TeXitoi (changes requested):
Thanks!

It lacks a few things:
* ChangeLog entry
* Documentation
* Maybe a relevant example
* A test with no bound, and an ignore field
* A test with a bound that is not StructOpt (FromStr seems a good candidat?)

structopt-derive/src/lib.rs
```diff
@@ -748,19 +748,115 @@ fn gen_paw_impl(name: &Ident) -> TokenStream {
...
+
+    if let Some(where_clause) = &generics.where_clause {
+        for predicate in &where_clause.predicates {
+            if let WherePredicate::Type(predicate) = predicate {
+                let predicate_bounded_ty = &predicate.bounded_ty;
+                if type_param_bounds_contains(&predicate.bounds, "StructOpt") {
+                    trait_bound_amendments.add(quote!{ #predicate_bounded_ty : ::structopt::StructOptInternal });
+                }
+                if type_param_bounds_contains(&predicate.bounds, "FromStr") {
+                    trait_bound_amendments.add(quote!{ < #predicate_bounded_ty as ::std::str::FromStr>::Err : ::std::fmt::Display + ::std::fmt::Debug });
```
TeXitoi, line 833: I don't really like this, as if we do this for FromStr, we will need to do it for TryFrom, From, and possibly more. Can't we just, for now, let the user put all the constraint himself? Is there a technical reason I don't see to this?

TeXitoi: If the user can put these bound manually, I propose we don't do this magic, and well see at usage, to choose if we use it later.

TeXitoi: because removing this would be a breaking change, but adding it is not (I think).

TeXitoi (approved):

TeXitoi:
Thanks!

TeXitoi:
v0.3.22 published

PR #485, opened 2021-07-18
Bump strum dev-dependency to latest version (0.20 -> 0.21)

decathorpe:
This requires no code changes, the example still compiles fine.

strum 0.21 claims to have an MSRV of 1.32.0, and structopt's MSRV is 1.36.0, so that should not be a problem either.


TeXitoi:
Thanks

PR #491, opened 2021-08-21
Fix the problem where the build fails due to the ambiguous type of `map`

### LINKED ISSUE #490: [E0282] Cannot infer the result type of `map`
If there is a type `T` for which `FromStr` and `FromIterator` are implemented and is used in an option as `Vec<T>` or `Option<Vec<T>>`, the build will fail by the following error.

```
error[E0282]: type annotations needed
   --> tests/issues.rs:140:18
    |
140 |         opt_vec: Vec<u16>,
    |         ---------^^^^^^^^
    |         |        |
    |         |        cannot infer type for type parameter `B` declared on the associated function `map`
    |         this method call resolves to `T`
```

The whole example code is

```rust
    struct U16ish;
    impl FromStr for U16ish {
        type Err = ();
        fn from_str(_: &str) -> Result<Self, Self::Err> {
            unimplemented!()
        }
    }
    impl<'a> FromIterator<&'a U16ish> for Vec<u16> {
        fn from_iter<T: IntoIterator<Item = &'a U16ish>>(_: T) -> Self {
            unimplemented!()
        }
    }

    #[derive(StructOpt, Debug)]
    struct Opt {
        opt_vec: Vec<u16>,
        #[structopt(long)]
        opt_opt_vec: Option<Vec<u16>>,
    }
```

This situation may sound unlikely at first, but it actually happens when you use Structopt with [Indicatif](https://github.com/mitsuhiko/indicatif) crate on Windows. Indicatif defines `Utf16Char` on Windows and it causes this error.

nullpo-head:
This PR closes #490. Please refer to #490 for the detail of the problem.
Let me know if you want to make `convert_type` a function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v0.3.23 (unreleased)
 
 * Update minimal rust version to 1.46 because of bitflags 1.3
+* Fixed [a bug that occurs when the type of `map` becomes ambiguous](https://github.com/TeXitoi/structopt/issues/490).
 
 # v0.3.22 (2021-07-04)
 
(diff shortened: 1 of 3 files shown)


nullpo-head:
@TeXitoi Thanks! I've updated the changelog.

TeXitoi:
Thanks

TeXitoi:
v0.3.23 published

PR #502, opened 2021-10-08
Update keyvalue.rs

Eolien55:
I updated `keyvalue.rs` to make it coherent with the actual behavior, since `number_of_values` is useless in that case.


TeXitoi:
I don't get it, it's clearly explained that, if you don't `number_of_values`, it will break positional argument after.

Issue #339, opened 2020-01-23
Q: How to re-export structopt from my crate?


cecton:
I have a lib that is using structopt and I would like to re-export it. Unfortunately it seems I can't do that because the macros are not reexported too:

In crate A:
```rust
pub use structopt::{StructOpt, clap};
```

In crate B:
```rust
use a::StructOpt;
```

But I get this compilation error:

```
   Compiling a v2.0.0 (...)
error[E0433]: failed to resolve: could not find `structopt` in `{{root}}`
  --> src/cli.rs:25:24
   |
25 | #[derive(Clone, Debug, StructOpt)]
   |                        ^^^^^^^^^ could not find `structopt` in `{{root}}`
```

CreepySkeleton:
This is one of the core problems with proc-macros. Structopt generates something like this:
```rust
::structopt::StructOpt::some_method(whatever)
```
You see? It assumes that your crate depends on `structopt`, e.g  `::structopt` is a valid path. 

It has to do that because there's no guarantee that you `use structopt::StructOpt as StructOpt;`, you can very well rename this import (and some actually do). The "full path" approach proved least erroneous.

As any heuristic, it has it's own drawbacks, one of them being that you can't simply reexport the trait.  This `B` crate has to depend on structopt too, this is the only way for now. (Keep in mind that the versions must match or [bad things will happen](https://github.com/TeXitoi/structopt/issues/315#issuecomment-568502792))

We would accept a PR that implements this serde-like approach (yes, serde has the very same problem):
```rust
#[derive(StructOpt)]
#[structopt(crate = a::structopt)] // explicitly provide the path for structopt
struct Opt {}
```

cecton:
Inactive, closing

kszela24:
Can we reopen this?  I'd like to contribute this sometime soon to make a crate I've developed work without requiring others to explicitly specify `structopt` as a dependency.  I'm running into the same issue mentioned here with: https://github.com/kszela24/clickrs

PR #506, opened 2021-10-18
Custom Structopt Crate Path

kszela24:
This PR provides the initial work around passing `structopt_path` to the procedural macro (https://github.com/TeXitoi/structopt/issues/339).

There's two places left that I'm having trouble seeing where I can pull the `structopt_path` attribute in:

https://github.com/TeXitoi/structopt/blob/da1fff81aded1c239ffcbd0a27ccdc7f28f74ff2/structopt-derive/src/lib.rs#L993-L1004

Not entirely sure how to handle this, I'm guessing the dummy returned by this block wouldn't be ideal for a user if it ended up getting used in a package built on top of structopt?

The other location is in attrs.rs:

https://github.com/TeXitoi/structopt/blob/da1fff81aded1c239ffcbd0a27ccdc7f28f74ff2/structopt-derive/src/attrs.rs#L322-L331

Additionally, I had tried using `crate` as serde does in their implementation but kept running into a compilation error: "expected identifier" (which I guess creates an issue because it's a keyword) so went with this naming.  Not sure what's best to name it if not `crate`,  so went with `structopt_path`.  It might be due to the way that serde is parsing the macro attributes but the way I've done it is consistent with the rest of the cases for parsing the arguments.  Happy to give it another go if you'd like this to be consistent with serde.

- [ ] Add unit tests
- [ ] Add documentation around extending structopt with the new functionality that explicitly mentions the “structopt_path” argument must come first in the arguments to the procedural macros.
- [ ] Try to get the one remaining case working based on the suggested solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v0.3.26 (2021-10-19)
+
+* Added ability to specify custom structopt crate path [#506](https://github.com/TeXitoi/structopt/pull/506)
+
 # v0.3.25 (2021-10-18)
 
 * Fix duplication of aliases in subcommands [#504](https://github.com/TeXitoi/structopt/pull/504)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v0.3.26 (2021-10-19)
+# v0.3.25 (2021-10-19)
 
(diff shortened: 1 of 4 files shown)


kszela24:
Also happy to add a small section in the documentation around extending structopt and using this new functionality

TeXitoi:
This is an enhancement, and structopt is now feature frozen. Please propose it to clap.

Thanks for your contribution!

PR #529, opened 2022-10-13
Add StructOpt::gen_help to generate help string

yanganto:
Hi there,
Thanks for providing this fantastic crate, I love it so much.

If there is a way to generate the help string from `struct` directly will be really helpful for me.

This method may be a good idea to do.
`<Opt as StructOpt>::gen_help()`

Such that I can avoid getting the help string by calling the program itself again like following
https://github.com/yanganto/s3rs/blob/master/src/command/mod.rs#L539-L547

If there is anything more I can do in this PR please kindly let me know.
Thank you for reviewing.

### DIFF
--- a/examples/gen_help.rs
+++ b/examples/gen_help.rs
@@ -1,0 +1,25 @@
+use structopt::StructOpt;
+
+/// A basic example
+#[derive(StructOpt, Debug)]
+#[structopt(name = "basic")]
+struct Opt {
+    /// Activate debug mode
+    #[structopt(short, long)]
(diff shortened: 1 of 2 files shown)


TeXitoi:
Hi.

StructOpt does not get new features. The complete functionalities are now present in clap since v3, so new features go on clap.

Also, note that you can implement this from outside structopt, so this PR will not solve something that's not solvable from the outside.

Thanks for the contribution.

yanganto:
Got it, thanks for the info.

PR #532, opened 2023-06-29
tests: fix overridable()

Lencerf:
Tested:
cargo test -p structopt --test rename_all_env


TeXitoi:
Thanks. Don't forget that structopt is in maintenance, you should now use clap.

PR #536, opened 2024-01-07
Replace proc-macro-error with proc-macro2-diagnostics

tamird:
See individual commits.

I realize this crate is deprecated in favor of clap, but it still seems to
enjoy some 10,000-40,000 downloads per day over the past three months. Updating
to syn@2 is likely to have positive effect on compilation time for dependents.

See https://crates.io/crates/structopt-derive for download graphs.

### DIFF
--- a/tests/ui/raw.stderr
+++ b/tests/ui/raw.stderr
@@ -1,7 +1,7 @@
 error: `#[structopt(raw(...))` attributes are removed in structopt 0.3, they are replaced with raw methods
 
-  = help: if you meant to call `clap::Arg::raw()` method you should use bool literal, like `raw(true)` or `raw(false)`
-  = note: if you need to call `clap::Arg/App::case_insensitive` method you can do it like this: #[structopt(case_insensitive = true)]
+         = help: if you meant to call `clap::Arg::raw()` method you should use bool literal, like `raw(true)` or `raw(false)`
+         = note: if you need to call `clap::Arg/App::case_insensitive` method you can do it like this: #[structopt(case_insensitive = true)]
 
   --> $DIR/raw.rs:13:17
    |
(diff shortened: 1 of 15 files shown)


tamird:
@TeXitoi could you have a look please?

TeXitoi:
That’s a big change for a frozen crate. I’ll try to have a look, but no promise.

tests/ui/raw.stderr
```diff
@@ -1,18 +1,10 @@
 error: `#[structopt(raw(...))` attributes are removed in structopt 0.3, they are replaced with raw methods
-
-  = help: if you meant to call `clap::Arg::raw()` method you should use bool literal, like `raw(true)` or `raw(false)`
-  = note: if you need to call `clap::Arg/App::case_insensitive` method you can do it like this: #[structopt(case_insensitive = true)]
```
TeXitoi, line 4: This is a regression

tamird: Yes, and one that I've called out [here](https://github.com/TeXitoi/structopt/pull/536/files#diff-65b84fea55a38ced0e2d71af5d960af930f9cf65279f00411af3e36fe673332aR131).

epage:
> I realize this crate is deprecated in favor of clap, but it still seems to
enjoy some 10,000-40,000 downloads per day over the past three months. Updating
to syn@2 is likely to have positive effect on compilation time for dependents.

imo that effort would be better spent migrating some of the more popular users.

tamird:
You mean like this?
https://github.com/rust-lang/git2-rs/pull/1007
https://github.com/brendanzab/codespan/pull/357
https://github.com/marshallpierce/rust-base64/pull/262

tests/ui/raw.stderr
jayvdb: That regression should be noted in a CHANGELOG.md entry ?

tamird: Would that unblock this PR?

TeXitoi: No regression allowed for a crate in maintenance just to optimize compilation time. Better to migrate to clap if you want to improve compilation times.

TeXitoi:
I close this MR as there is no activity.

tamird:
If you can close it, maybe you can also review and accept it?

TeXitoi:
Sorry. I don't want a regression, and there is one.

Also, I'm quite anxious at this PR as this is a big library change, and it might cause a behavior change.

I know that you worked in this and that closing this deny your work, but it adds more risk than advantages for my point of view.

tamird:
Ack. Salvaged some useful stuff into #542 + added GitHub CI.

PR #542, opened 2024-09-30
Rehabilitate CI, upgrade to syn@2

tamird:
See individual commits. Extracted from #536 minus the regression.


tamird:
@TeXitoi I think you might need to enable github actions from the repo settings.

TeXitoi:
Sorry for not responding in so much time.

I think syn upgrade, while meaningful, is too risky in term of back compatibility. If you want to avoid depending on syn v1, better to upgrade to clap_derive.

Also, this paw thing might be a breaking change.

The github action is quite nice, but regarding the number of MR I have, I think that’s not really critical.

